PR #38, opened 2014-08-14
Fix installation paths.

dominikgrewe:
When installing cutorch locally, the installer tries to copy THC
files outside of the rocks tree (e.g. /usr/lib/libTHC.so or
/usr/include/THC/THC.c). Consequently, a local installation fails
because the user isn't allowed to copy files to /usr.
This patch fixes the installation paths so everything is copied into
rocks trees only, thus fixing the local installation problem.


soumith:
we made local installs the default setting for EVERYONE. so this issue is no longer relevant.

dominikgrewe:
@soumith Sorry to pick this up again, but what do you mean by "we made local installs the default setting for everyone"? I still can't install cutorch locally with luarocks, i.e. `luarocks install --local cutorch` fails with
```bash
CMake Error at lib/THC/cmake_install.cmake:42 (FILE):
  file INSTALL cannot copy file
  "/tmp/luarocks_cutorch-scm-1-7185/cutorch/build/lib/THC/libTHC.so" to
  "/usr/lib/libTHC.so".
```

soumith:
if you installled torch to ~/torch/install then libTHC.so should be attempted to be installed at ~/torch/install/lib/

PR #84, opened 2014-11-20
Move global cublas state into cutorch table.

### LINKED ISSUE #80: Moving static cuBlas state into THCudaState
Follow-up from @clementfarabet's patch. @soumith, did you say you wanted to do this? If not, I'm happy to write a patch. Just wanted to check that you haven't got one already.

dominikgrewe:
Create new THCudaBlasState struct to hold the state and add it to
THCudaState.
Add new utils.h/c with getState function

Fixes #80

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -39,7 +39,7 @@ ADD_SUBDIRECTORY(lib)
 INCLUDE_DIRECTORIES("${CMAKE_CURRENT_SOURCE_DIR}/lib/THC")
 INCLUDE_DIRECTORIES("${CMAKE_CURRENT_SOURCE_DIR}/torch")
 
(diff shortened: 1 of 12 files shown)


dominikgrewe:
Just thought I'd give you an update on this, to make sure it's going in the right direction:

In order to pass a state to every THC function we can't use the macro-based approach any more where we simply include files like "generic/THTensor.h" etc. So I've copied an instantiated version (Real = Cuda etc) of those files into cutorch.
Every function will need a state as the first argument now, so `THCudaTensor_nElement(t)` becomes `THCudaTensor_nElement(state, t)` etc.

I'll create a new PR when I'm done. If you have any comments in the meantime, please let me know.

soumith:
fixed via #103

PR #88, opened 2014-12-06
Supporting functions for Complex Convolutions

jonathantompson:
added cgemv and cgemm methods.

Note: I didn't want to add a cgemv and cgemm CPU implementation (in the torch repo) because I saw no easy way to add complex number support on the CPU side (without a huge amount of work).  Likewise, I don't need it and I'm super busy, so I didn't have time to implement it.  So instead I've just added 2 extra methods to the THCBlas.h header...  I don't "love" that I added extra methods into the header (since the header is autogenerated from the CPU counterpart), but hopefully code reviewers aren't going to be too offended.

The alternative (that Soumith pointed out to me), would be to add these methods to SpatialConvolutionComplexMM.cu and keep them out of THCBlas.  Personally, I think the should go here because someone might want them later down the line.

### DIFF
--- a/lib/THC/THCBlas.cu
+++ b/lib/THC/THCBlas.cu
@@ -165,6 +165,35 @@ void THCudaBlas_gemv(char trans, long m, long n, float alpha, float *a, long lda
           "in the range 0 < [val] <= %d", INT_MAX);
 }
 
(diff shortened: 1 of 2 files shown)


soumith:
https://www.youtube.com/watch?v=dQw4w9WgXcQ

Issue #91, opened 2014-12-13
cutorch var() and std() don't allow dimension argument


szagoruyko:
Incompatibility with torch. To reproduce:

```lua
r = torch.rand(8,2)
print(r:var(1))  -- OK
print(r:cuda():var(1))    -- not OK
print(r:cuda():std(1))    -- not OK
print(r:cuda():mean(1))  -- OK!
```

soumith:
fixed now in trunk via #96

Issue #70, opened 2014-11-06
missing API against TH


soumith:
The following math functions are missing in THC but present in TH:
- [x] numel
- [x] prod
- [x] zeros
- [x] ones
- [x] reshape
- [x] rand
- [x] randn
- [x] round
- [x] atan2
- [x] std (not stdall, this is per-dimension)
- [x] var (not varall, this is per-dimension)
- [x] cumsum
- [x] cumprod
- [x] maskedFill
- [x] maskedSelect
- [x] sort
- [x] maskedCopy (waiting for merge, https://github.com/torch/cutorch/pull/167)
- [x] multinomial (implemented, waiting for PR)
- [x] cross
- [x] logicalall
- [x] logicalany
- [x] tril
- [x] triu
- [x] trace
- [x] diag
- [x] nonzero
- [x] range
- [x] cat
- [x] linspace
- [x] logspace
- [ ] eye
- [ ] randperm
- [ ] histc
- [ ] conv2 (these are torch operators, not to be confused with nn Convolutions which are already implemented)
- [ ] conv3

When these are implemented, cwrap entries can be added that would make cutorch completely API compatible with torch

dominikgrewe:
Does anyone have an implementation of std and var yet? Otherwise I'll have a go next week.
Afaict, we can't use `THCudaTensor_reduceDim` so we'll need something custom (but with a similar structure as reduceDim).

soumith:
no, do not have it. this has been long overdue, but we should co-ordinate these, let me email you guys.

dominikgrewe:
Any progress on cumsum and cumprod?

soumith:
have not started them yet!

dominikgrewe:
Okay, I might have a go soon then, because we'd like to use it. Just waiting for the THCState PR to be merged.

soumith:
I will merge the THCState PR on Friday. All the patches have been prepared except for fbcunn, working on that as well.

dominikgrewe:
@soumith For reductions along a single dimension, we still have the restriction that the tensor must not have more than 4 dimensions. Did you say you're working on a fix for that? If so, what's the progress on it?
If you're not working on it, we could easily change the code to what we do for std and var, where there's no restriction on dimensionality at all.

soumith:
@wickedfoo already has  PR for that internally. Its all implemented. I'm on
vacay till 26th, so I will sync those changes at the end of this month.

soumith:
[...]

PR #96, opened 2015-01-06
Single-dimension std and var.

dominikgrewe:
Implementation of std and var across a single dimension. The computation
is the same as in Torch. The reduction code is similar to the one used
for sum, max etc., but it's not restricted to <= 4 dimensions.


soumith:
awesome! fixes #91 fixes one of the case in #70

PR #99, opened 2015-01-12
Pass a state to every THC function.

dominikgrewe:
Every THC function gets a THCState pointer as the first argument.
Some generic files that were previously included have been instantiated
because TH functions currently don't get a state parameter.


dominikgrewe:
This is relying on a patch to cwrap that adds a "registerDefaultArgument" method: https://github.com/torch/cwrap/pull/3
It makes passing the same argument to each function a lot simpler. Let me know if anyone can think of a better way of doing that.

soumith:
great! overall looks good. but i cannot ship this without a patch being prepared for cunn (as it will stop building).

soumith:
to make a list of what need to be patched with this change:
- [ ] cunn
- [ ] ccn2.torch (me)
- [ ] imagine-nn (@szagoruyko)
- [ ] cunnx (@nicholas-leonard)
- [ ] nnbhwd (@qassemoquab)

soumith:
to give a context to the others, this patch makes cutorch thread-safe, so you can start doing stuff like 1 thread per GPU

dominikgrewe:
Thanks Soumith. I've created a patch for cunn: torch/cunn#59

@koraykv also pointed out that I hadn't added the new .h files to the CMake file in lib/THC. This is now fixed.

soumith:
thanks for the cunn patch. i would not be able to finish the ccn2 patch till next week.

koraykv:
Should we wait for other projects to catch up or merge now?

soumith:
i would say wait, because these packages are popular and lots of people use them. it will be like introducing a build-break. i will try to get the patches out for the rest by mid next week.

koraykv:
Thanks!

szagoruyko:
@dominikgrewe have you got an example how to pass THCState to ffi?

dominikgrewe:
No, I don't have an example for ffi unfortunately.

From Soumith's earlier post I assume you're talking about the use of ffi in imagine-nn. From what I can tell you're wrapping all ffi calls in Lua, right? For example, Lua function `SpatialMaxPooling:updateOutput` calls C function `SpatialMaxPooling_updateOutput`. So you could add the state argument to the ffi functions, load the state in your Lua code (see my nn patch for an example) and pass that state to the ffi function. Would that work?

soumith:
szagoruyko, it's simple, just pass in cutorch._state to the ffi function in the first argument

szagoruyko:
Thanks, imagine-nn is ready https://github.com/szagoruyko/imagine-nn/tree/state

soumith:
had to make a small addition to this PR to support FFI: 
https://github.com/torch/cutorch/tree/THCState

dominikgrewe:
Thanks for the fix Soumith.

nicholas-leonard:
@soumith you can check the box for cunnx.

soumith:
ok fbcunn also ready, retested with new changes to cunn in this short period and made fixes.
merging now the ones within my power now.

soumith:
Boom! done. @szagoruyko and @nicholas-leonard merge your repos as soon as you get to it.

dominikgrewe:
Thanks! :)

szagoruyko:
@soumith it seems to me that cb12f6bf38ed45b05dea2c1930dcb718ef21f82a was left unmerged

soumith:
@szagoruyko thanks for spotting it, merged. merging 4 PRs at once had it's side effects.

PR #109, opened 2015-01-29
Add missing header file

santazhang:
nvcc is complaining: THCTensorConv.cu(340): error: identifier "THCudaTensor_copy" is undefined


soumith:
thanks

Issue #110, opened 2015-01-30
torch.cmul inconsistent behavior, mutates first argument?


louissmit:
Maybe im missing something, but in vanilla Torch z=torch.cmul(x,y) does not mutate x, while in cutorch it does:

```lua
require 'cutorch'

cuda = true

local x = torch.Tensor(4):fill(1)
local y = torch.Tensor(4):zero()

print(x)
if cuda then
    x = x:cuda()
    y = y:cuda()
end
local z = torch.cmul(x, y)
print(x)
>>>

 1
 1
 1
 1
[torch.FloatTensor of dimension 4]

0
0
0
0
[torch.CudaTensor of dimension 4]
```

soumith:
Thanks for the bug report. Fixed in trunk via commit: https://github.com/torch/cutorch/commit/0b6d71bf892ed245c42b8100d625401bf419d4e2

To get the fixes, reinstall cutorch with:
luarocks install cutorch

louissmit:
That was quick! Thanks, much appreciated!

PR #111, opened 2015-02-01
bugfix for pow and tpow for non-inplace operations

fmassa:
Same problem as https://github.com/torch/cutorch/issues/110 but for pow and tpow.


soumith:
Thanks

PR #112, opened 2015-02-05
Ignore cudaErrorPeerAccessAlreadyEnabled.

dominikgrewe:
If we call cudaDeviceEnablePeerAccess for the same devices twice, it
will return cudaErrorPeerAccessAlreadyEnabled. THCudaCheck treats that
as an error and fails, whereas it really only indicates that peer access
has already been enabled. In which case we shouldn't fail.


soumith:
:+1:

PR #115, opened 2015-02-25
Implementation of reduction without any limit in the number of dimensions

dominikgrewe:
Follows structure of the kernels to compute std & var.


soumith:
nice!

Issue #97, opened 2015-01-09
Repeated calls to uniform() produce non-uniform output


mdenil:
Repeated calls to `:uniform(0, 1)` produce non-uniform output.  This only happens with CudaTensors and only across many calls to uniform, filling a large tensor with a single call seems okay.

```
-- this produces uniform output
-- local block = 10000

-- this does not
local block = 1

local r = torch.zeros(block):cuda()
for j = 1, 10000/block do
    r:uniform(0, 1)
    -- this is also broken
    -- r:rand(r:size())

    for i = 1, r:size(1) do
        print(r[i])
    end
end
```

dominikgrewe:
Yes, you're right. The output seems to be "less uniform" when not generating a whole batch. It seems to get better the more numbers you generate but it's still not great.

I'm guessing that CUDA's random number generator isn't really designed for generating single numbers. To be efficient, you want to generate lots of them at a time. This is opposed to the Torch RNG where it doesn't make a difference whether you generate a single number N times, or N numbers in a batch (both in terms of performance and the result).

I don't think this is a problem in practice. If you only want to generate single numbers at a time, you can just use torch.uniform().

dominikgrewe:
We did some more digging into this:

It turns out that the numbers generated by our RNG are only properly distributed when generating at least a few thousand numbers at a time (we used the dieharder tests to verify that). This is only since we changed from using cuRAND's host API to its device API. The host API always generates good numbers.

Some profiling suggests that the host API is always generating 16384 (64*256) random numbers at a time and then copies numbers from that pool into the output buffer. I can think of two ways we could do something similar:

1) Generate a pool of random bits and transform them into the right distributions on-demand (this is what the host API is doing afaict).
2) Have different pools for different distributions (eg. uniform and normal) and transform those numbers to the right distribution on-demand.

While the second option may obviously use more memory than the first, it would probably be easier to implement because we don't have to think about how to transform random bits into uniformly (or normally) distributed floats, for example.

soumith:
this was documented in NVIDIA's CUDA docs, or I remember reading so.

Second option might end up using significantly more memory. If the overhead is less than say 10-25MB and it would save you tons of time, go for option 2. If the overhead of going for option (2) will hurt people with 2GB or 4GB cards (for example), then it is not great, and option (1) might be the way.

dominikgrewe:
[...]

PR #116, opened 2015-03-05
Fix RNG issues.

dominikgrewe:
Reset MTGP32 constants when setting the seed.
Always generate multiples of 256 numbers (discarding the ones not
needed).


dominikgrewe:
Fixes problem as described in torch/torch7#167

jucor:
Great fix Dominik, thanks! For the record: the only actual downside is very minor and not worth worrying about. In the worst case the period of the generator is divided by 256, but since MGTP32 has period  2^11213-1, dividing it by 256 makes it way long enough.

dominikgrewe:
To give some context to Julien's comment:
This fix also solves the issue discussed in #97 . I've tested the RNG using the dieharder test suite on all tests that accept floats as inputs. Whereas before all tests started failing when only generating 10s or 100s of numbers, they now succeed even when only generating single numbers at a time.

soumith:
that is great. thanks @dominikgrewe

PR #117, opened 2015-03-17
Expose permute operation for CudaTensors

dominikgrewe:


dominikgrewe:
Depends on torch/torch7#173.

PR #120, opened 2015-03-20
adds Sort, revamps TensorMath, adds masked* operations

soumith:
Implements for #70 :
* maskedSelect
* maskedCopy
* maskedFill
* sort

Adds a TensorApply and TensorReduce kernels that work on both contiguous and non-contiguous tensors, and maintains equal performance for contiguous cases.

This PR makes performance with non-contiguous kernels a huge huge improvement.

Implemented completely by @wickedfoo

Having some build issues, looking into it


soumith:
Also has a multi-GPU copy performance enhancement by @adamlerer

wickedfoo:
also fyi re: @adamlerer 's change, that's actually a bug fix. both the old and new non-contiguous copy kernels were missing the same synchronization semantics implied by cudaMemcpyAsync() when copying between GPUs; cudaMemcpyAsync() actually synchronizes in both src and dst default streams, whereas a kernel launch only gets inserted into the stream of the GPU that it is running on.

dominikgrewe:
Thanks guys. That looks really impressive!
Can we leave this up for a couple of days before merging so people can digest it? :)

I just have two questions from having a first glance:

1) For apply, you fix the number of elements processed per thread to 8 so the compiler can unroll the loop. When processing small tensors, that means we probably don't fully utilise the GPU. Have you looked into that?

2) Functions such as `THCudaTensor_pointwiseApply1` return false for some errors and call THError for others. When just returning false the user doesn't really know why something went wrong. Is it better to just stick to calling THError in all cases?

wickedfoo:
1. For larger tensors it better amortizes the initial setup cost, and I
tried 1-16 in powers of 2 and 8 was about optimal.

I believe when I looked at the SASS while working on this (and on other
codes) the compiler interleaves non-interdependent loads to some degree
based on its operation timing heuristics in the unrolled loop, so that
multiple loads can be in flight at once, so there may not be much benefit
to having more threads perform the load transactions that a smaller number
of threads would be doing concurrently anyways. Increased ILP through fewer
interdependencies and more in-flight operations versus greater TLP with
fewer in-flight operations per thread, though I may not actually be using
the ILP by not giving it temporary storage for the loads.

For really small tensors, note that if the tensor has less than 4096
elements, it will only be processed by one block (thus blockId ==
getLastLinearBlockId<IndexType>()), which doesn't use the unrolled path.
There, each thread issues one load and loops in a block-wide stride. In
that case, there could be a perf hole between 512-4096 elements, since I
could have more warps active.

I will look into 1-16384ish element tensors, inspect the SASS again and do
more detailed analysis.

2. Sounds good, though in some cases the caller may have allocated memory
which will not be properly cleaned up upon that error (because there's no
RAII style allocation, instead depending upon THCudaTensor_free() etc, like
the error I pointed out). The pointwise functions allocate no memory,
leaving that response to the caller. If anything I'd argue that the cuda
error check should also be moved out so the only failure the pointwise
functions have is if the tensor is too large or has too many dimensions,
which is returned via true/false.

wickedfoo:
Oh, and another comment about sort(). Some caveats about the implementation:

-Handles sort along any dimension (contiguous or not) if the size of each slice is 1-2048 elements. Thus, a sort(2) on a CudaTensor(128, 128, 128, 128) will work.
-Handles sort along contiguous dimensions if the total number of slices is less than 16, regardless of the size of the slice. Thus, a sort(2) on a CudaTensor(3, 100000000) will work. This is delegated to (number of slices) Thrust kernel launches.
-What is not yet implemented are cases not falling into the above two categories. Thus, a sort(2) on a CudaTensor(16, 5000, 16, 16) will not work (non-contiguous, >2048 elements), or a sort(2) on a CudaTensor(500, 5000) will not work (contiguous, more than 500 slices, each slice larger than 2048).

I set that aside to work on other things. That can be fixed sooner or later if important.

wickedfoo:
*nb: pointwise functions allocate no memory except in the exceedingly weird case where you are writing into a tensor with multiple indices that map to the same element, say an unfold-ed tensor, which I argue should just be an error in Torch (unless you are updating just a single element). cunn had a test for this.

Instead, I follow the old behavior is that both arguments are made contiguous, the operation is made, and then the result is written back into the overlapping index tensor using the copy kernel again. The only reason why you got a result that you expected before is because of several properties:
-there’s a write race between multiple GPU threads to write into the one destination memory place;
-an arbitrary choice (dependent upon kernel execution order, etc.) is made as to who wins;
-writing a float in CUDA via a data race does not result in tearing (e.g., the lower half word comes from one thread, the upper half comes from another).

In case where all source values are the same for a single output, you get what you expect; in the case where values are different, it depends upon the implementation and execution order of the non-contiguous copy kernel.

nicholas-leonard:
This PR only seems to have unit tests for non-contiguous multi-gpu copy. It should include tests for masked*, sort and most of the other use cases that were added.

wickedfoo:
Perf relative to the old Thrust implementation for 1-d contiguous tensors. apply1 here is add(1.0), apply2 is a:add(b), apply3 is torch.cmul(c, a, b). New code is faster for large tensors, but slower for small tensors, to a larger degree than desired (but still everything is in the us range, and typical noise is 1-5 us anyways). I'll work on perf for small tensors.

operation | 1-d size | new time in us (K40) | old time in us (K40) | relative speedup/slowdown

apply1 1 3.616000 2.559000 0.707688x
apply1 2 2.144000 1.856000 0.865672x
apply1 4 1.984000 1.696000 0.854839x
apply1 8 1.984000 1.728000 0.870968x
apply1 16 1.984000 1.728000 0.870968x
apply1 32 1.952000 1.632000 0.836066x
apply1 64 1.952000 1.632000 0.836066x
apply1 128 2.016000 1.728000 0.857143x
apply1 256 2.048000 1.728000 0.84375x
apply1 512 2.080000 1.760000 0.846154x
apply1 1024 2.336000 1.920000 0.821918x
apply1 2048 2.912000 2.848000 0.978022x
apply1 4096 4.095000 3.008000 0.734554x
apply1 8192 5.408000 2.464000 0.455621x
apply1 16384 4.992000 2.144000 0.429487x
apply1 32768 4.831000 2.432000 0.503415x
apply1 65536 4.096000 2.848000 0.695312x
apply1 131072 4.672000 3.616000 0.773973x
apply1 262144 8.671000 5.535000 0.638335x
apply1 524288 21.53500 19.68000 0.913861x
apply1 1048576 38.46300 37.69500 0.980033x
apply1 2097152 72.92600 73.27900 1.00484x
apply1 4194304 144.4120 145.1170 1.00488x
apply1 8388608 284.1530 289.2410 1.01791x
apply1 16777216 565.0740 578.7060 1.02412x
apply1 33554432 1.12e+03 1.16e+03 1.03571x
apply1 67108864 2.25e+03 2.35e+03 1.04444x
apply1 134217728 4.48e+03 4.78e+03 1.06696x
apply2 1 3.936000 2.368000 0.601626x
apply2 2 2.176000 1.792000 0.823529x
apply2 4 2.080000 1.696000 0.815385x
apply2 8 2.080000 1.728000 0.830769x
apply2 16 2.048000 1.728000 0.84375x
apply2 32 2.048000 1.728000 0.84375x
apply2 64 2.080000 1.728000 0.830769x
apply2 128 2.080000 1.760000 0.846154x
apply2 256 2.080000 1.760000 0.846154x
apply2 512 2.208000 2.688000 1.21739x
apply2 1024 2.528000 2.304000 0.911392x
apply2 2048 3.232000 2.304000 0.712871x
apply2 4096 4.448000 2.400000 0.539568x
apply2 8192 5.760000 1.952000 0.338889x
apply2 16384 5.247000 2.144000 0.408614x
apply2 32768 5.184000 2.848000 0.549383x
apply2 65536 4.768000 3.968000 0.832215x
apply2 131072 5.952000 6.527000 1.09661x
apply2 262144 14.68800 14.52700 0.989039x
apply2 524288 31.19900 29.63100 0.949742x
apply2 1048576 58.11000 58.20700 1.00167x
apply2 2097152 109.3410 113.3090 1.03629x
apply2 4194304 215.0030 227.7380 1.05923x
apply2 8388608 424.1500 450.6770 1.06254x
apply2 16777216 853.3230 905.5130 1.06116x
apply2 33554432 1.71e+03 1.78e+03 1.04094x
apply2 67108864 3.41e+03 3.57e+03 1.04692x
apply2 134217728 6.81e+03 7.13e+03 1.04699x
apply3 1 3.968000 2.848000 0.717742x
apply3 2 2.432000 1.888000 0.776316x
apply3 4 2.368000 1.856000 0.783784x
apply3 8 2.304000 1.727000 0.749566x
apply3 16 2.272000 1.760000 0.774648x
apply3 32 2.272000 1.728000 0.760563x
apply3 64 2.304000 1.728000 0.75x
apply3 128 2.272000 1.760000 0.774648x
apply3 256 2.336000 1.824000 0.780822x
apply3 512 2.368000 1.920000 0.810811x
apply3 1024 2.687000 2.144000 0.797916x
apply3 2048 3.360000 3.264000 0.971429x
apply3 4096 4.576000 3.296000 0.72028x
apply3 8192 5.920000 2.720000 0.459459x
apply3 16384 5.504000 2.496000 0.453488x
apply3 32768 5.440000 2.911000 0.53511x
apply3 65536 4.896000 3.712000 0.75817x
apply3 131072 7.968000 7.648000 0.959839x
apply3 262144 19.26300 16.19100 0.840523x
apply3 524288 31.48800 29.02300 0.921716x
apply3 1048576 58.55900 55.80700 0.953005x
apply3 2097152 111.6460 109.3100 0.979077x
apply3 4194304 217.2750 222.0110 1.0218x
apply3 8388608 429.5570 431.3490 1.00417x
apply3 16777216 945.2560 867.2420 0.917468x
apply3 33554432 1.71e+03 1.77e+03 1.03509x
apply3 67108864 3.44e+03 3.63e+03 1.05523x
apply3 134217728 6.83e+03 7.61e+03 1.1142x

Perf relative to old newContiguous/Thrust/copy to non-contig/free code for non-contiguous 2-d tensors with transpose(1, 2) applied to all arguments. Same operations as above. (Though, as an optimization, I could detect that for apply1 the layout is contiguous after some permutation, and for apply2/3 all arguments have the same size/stride arrays and is contiguous after some permutation, and do the application in the reordered view, so this would map to the contiguous case, but this currently forces the non-contiguous path for both old and new code).

operation | 2-d size | new time in us (K40) | old time in us (K40) | relative speedup/slowdown

apply1_noncontig 2x2 34.093856811523 264.16778564453 7.74825x
apply1_noncontig 4x4 31.948089599609 188.82751464844 5.91045x
apply1_noncontig 8x8 23.126602172852 168.0850982666 7.26804x
apply1_noncontig 16x16 23.126602172852 126.12342834473 5.45361x
apply1_noncontig 32x32 30.994415283203 166.89300537109 5.38462x
apply1_noncontig 64x64 40.054321289062 148.05793762207 3.69643x
apply1_noncontig 128x128 42.915344238281 135.18333435059 3.15x
apply1_noncontig 256x256 55.074691772461 148.05793762207 2.68831x
apply1_noncontig 512x512 107.0499420166 1950.0255584717 18.216x
apply1_noncontig 1024x1024 334.02442932129 1926.8989562988 5.76874x
apply1_noncontig 2048x2048 1431.941986084 2746.1051940918 1.91775x
apply1_noncontig 4096x4096 7370.9487915039 10210.037231445 1.38517x
apply1_noncontig 8192x8192 50721.168518066 72425.842285156 1.42792x
apply2_noncontig 2x2 42.200088500977 211.95411682129 5.0226x
apply2_noncontig 4x4 22.88818359375 190.97328186035 8.34375x
apply2_noncontig 8x8 23.126602172852 172.13821411133 7.4433x
apply2_noncontig 16x16 23.126602172852 171.89979553223 7.43299x
apply2_noncontig 32x32 48.87580871582 159.97886657715 3.27317x
apply2_noncontig 64x64 48.87580871582 149.96528625488 3.06829x
apply2_noncontig 128x128 55.074691772461 156.87942504883 2.84848x
apply2_noncontig 256x256 74.148178100586 926.97143554688 12.5016x
apply2_noncontig 512x512 160.93254089355 1886.1293792725 11.72x
apply2_noncontig 1024x1024 740.05126953125 1858.9496612549 2.51192x
apply2_noncontig 2048x2048 3013.8492584229 2702.9514312744 0.896844x
apply2_noncontig 4096x4096 12011.051177979 10941.982269287 0.910993x
apply2_noncontig 8192x8192 81535.816192627 75384.140014648 0.924552x
apply3_noncontig 2x2 36.001205444336 272.98927307129 7.58278x
apply3_noncontig 4x4 23.841857910156 207.90100097656 8.72x
apply3_noncontig 8x8 36.001205444336 248.90899658203 6.91391x
apply3_noncontig 16x16 33.140182495117 180.95970153809 5.46043x
apply3_noncontig 32x32 44.822692871094 171.18453979492 3.81915x
apply3_noncontig 64x64 50.067901611328 168.0850982666 3.35714x
apply3_noncontig 128x128 55.074691772461 202.89421081543 3.68398x
apply3_noncontig 256x256 77.009201049805 1084.0892791748 14.0774x
apply3_noncontig 512x512 216.00723266602 2681.9705963135 12.4161x
apply3_noncontig 1024x1024 936.98501586914 2667.9039001465 2.84733x
apply3_noncontig 2048x2048 2810.001373291 3664.0167236328 1.30392x
apply3_noncontig 4096x4096 12379.884719849 13031.959533691 1.05267x
apply3_noncontig 8192x8192 82146.883010864 89496.13571167 1.08946x

lib/THC/THCTensorSort.h
```diff
@@ -0,0 +1,12 @@
+#ifndef TH_CUDA_TENSOR_SORT_INC
+#define TH_CUDA_TENSOR_SORT_INC
+
+#include "deeplearning/torch/cutorch/lib/THC/THCTensor.h"
```
dominikgrewe: Fix path.

lib/THC/THCApply.cuh
```diff
@@ -0,0 +1,694 @@
...
+}
+
+#undef THC_APPLY_ELEM_PER_THREAD
+#undef THC_APPLY_THREADS_PER_BLOCK
+
+// Implementation of copyIgnoringOverlaps, defined after pointwiseApply2.
+void THCudaTensor_copyIgnoringOverlaps(THCState* state,
+                                       THCudaTensor* dst,
+                                       THCudaTensor* src) {
+THCudaTensor_pointwiseApply2(state, dst, src, CopyOp<float>(),
```
dominikgrewe: Fix indentation

lib/THC/THCReduce.cuh
```diff
@@ -0,0 +1,319 @@
...
+  // Reduce within the block
+  extern __shared__ float smem[];
+  smem[threadIdx.x] = r;
+
+  // First warp will perform reductions across warps
+  __syncthreads();
+  if ((threadIdx.x / 32) == 0) {
+    r = init;
+    for (int i = 0; i < blockDim.x; i += 32) {
+      r = reduceOp(r, modifyOp(smem[i + threadIdx.x]));
```
dominikgrewe: I don't think we want to modify the intermediate results, do we? Same in the loop below.

lib/THC/THCReduce.cuh
```diff
@@ -0,0 +1,319 @@
...
+  }
+
+  // First thread will perform reductions across the block
+  __syncthreads();
+  if (threadIdx.x == 0) {
+    r = init;
+#pragma unroll
+    for (int i = 0; i < 32; ++i) {
+      r = reduceOp(r, modifyOp(smem[i]));
+    }
```
dominikgrewe: Are these two loops faster than doing a tree-style reduction? I can see why this might be the case (less synchronization and loop unrolling). I'm just curious if you benchmarked it.

lib/THC/THCReduce.cuh
wickedfoo: What do you mean by modify the intermediate results? Do you mean using a separate variable for each level of reduction to be clearer:
```
float threadWide = init;
...
smem[threadIdx.x] = threadWide;
__syncthreads();
...
float warpWide = ...;
...
smem[laneId] = warpWide;
__syncthreads();
...
float blockWide = init;
...
```

?

dominikgrewe: You're currently calling modifyOp on the intermediate results. That doesn't look right to me.

wickedfoo: ah, that, yes that's wrong.

lib/THC/THCReduce.cuh
wickedfoo: It's up to 70% or so faster:
```
369.78ms  2.8480us              (1 1 1)       (512 1 1)        12  2.0480KB        0B         -           -   Tesla K40m (0)         1         7  reduceByWarp(void) [352]
369.80ms  2.9440us              (4 1 1)       (512 1 1)        12  2.0480KB        0B         -           -   Tesla K40m (0)         1         7  reduceByWarp(void) [354]
369.81ms  2.9440us             (16 1 1)       (512 1 1)        12  2.0480KB        0B         -           -   Tesla K40m (0)         1         7  reduceByWarp(void) [356]
369.83ms  4.6400us             (64 1 1)       (512 1 1)        12  2.0480KB        0B         -           -   Tesla K40m (0)         1         7  reduceByWarp(void) [358]
369.84ms  10.272us            (256 1 1)       (512 1 1)        12  2.0480KB        0B         -           -   Tesla K40m (0)         1         7  reduceByWarp(void) [360]
369.85ms  34.560us           (1024 1 1)       (512 1 1)        12  2.0480KB        0B         -           -   Tesla K40m (0)         1         7  reduceByWarp(void) [362]
369.89ms  131.71us           (4096 1 1)       (512 1 1)        12  2.0480KB        0B         -           -   Tesla K40m (0)         1         7  reduceByWarp(void) [364]
370.03ms  3.1680us              (1 1 1)       (512 1 1)         8  2.0480KB        0B         -           -   Tesla K40m (0)         1         7  reduceTree(void) [366]
370.03ms  3.2960us              (4 1 1)       (512 1 1)         8  2.0480KB        0B         -           -   Tesla K40m (0)         1         7  reduceTree(void) [368]
370.04ms  3.4240us             (16 1 1)       (512 1 1)         8  2.0480KB        0B         -           -   Tesla K40m (0)         1         7  reduceTree(void) [370]
370.04ms  5.8560us             (64 1 1)       (512 1 1)         8  2.0480KB        0B         -           -   Tesla K40m (0)         1         7  reduceTree(void) [372]
370.05ms  16.096us            (256 1 1)       (512 1 1)         8  2.0480KB        0B         -           -   Tesla K40m (0)         1         7  reduceTree(void) [374]
370.07ms  57.535us           (1024 1 1)       (512 1 1)         8  2.0480KB        0B         -           -   Tesla K40m (0)         1         7  reduceTree(void) [376]
370.13ms  223.23us           (4096 1 1)       (512 1 1)         8  2.0480KB        0B         -           -   Tesla K40m (0)         1         7  reduceTree(void) [378]
```

wickedfoo: Faster still if I use __shfl_xor for the last level of reduction (go from one thread -> one warp) for CC 3.0+.

dominikgrewe: Cool, that's impressive! What's the difference between the first and second column here?

And the code would still be correct if the warp size wasn't 32, right? We only rely on it for performance, not correctness, if I read the code correctly.

wickedfoo: Everything except line 122 would work, but warpSize in CUDA is not a compile-time constant (maps to %WARP_SZ in PTX) so it can't unroll that loop if I used warpSize instead of 32. It depends upon all lanes in the first warp (of which there are 32) writing out values. 

In our own internal code we have headers that have all compute capability specific values like this defined as constants so they can affect loop unrolling (e.g., lanes per warp, max warps per block, ...). What I should do here is check against compute capability via `__CUDA_ARCH__` and error out if it's one that is unhandled perhaps (so for known __CUDA_ARCH__ values #define WARP_SIZE to 32 otherwise #error unhandled compute capability).

Usually I use a mix of shared memory and warp-wide __shfl_xor reductions when writing this kind of code internally. I can add the CC >= 3.0 __shfl_xor version. I'll play around to see if doing this by 16 or 8 is faster than 32.

First column and second column? That's nvprof --print-gpu-trace output. First col is just wall clock time (from program start), second is per-kernel timing.

dominikgrewe: What's wrong with line 122? It's only run on thread 0, so we're not relying on threads in a warp being synced. My main concern is that we don't rely on a warp size of 32 for correctness, and I don't think that's the case here.

wickedfoo: Ah sorry, you're right, not important for correctness. This code should be correct regardless of warp size as is current.

I missed the `if ((threadIdx.x / 32) == 0)` above, internally when I write code like this I use a `if (getWarpId() == 0)` which returns a linearized warp ID for the current thread, which works regardless of the CTA dimensions. In this case if the warp size is 16 or 48 or whatever, the sizes won't match up.

However, the `__syncthreads();` at line 118 can also be removed, since a warp sees its own writes. In this case, it would work only if `(threadIdx.x / 32) == 0` will be true for one warp only; i.e., something like:

```
 // First warp will perform reductions across warps
+  __syncthreads();
+  if ((threadIdx.x / 32) == 0) {
+    r = init;
+    for (int i = 0; i < blockDim.x; i += 32) {
+      r = reduceOp(r, modifyOp(smem[i + threadIdx.x]));
+    }
+
+    // Each lane participating writes out a value
+    smem[threadIdx.x] = r;
+
+    // First thread will perform reductions across the block
+    if (threadIdx.x == 0) {
+      r = init;
+#pragma unroll
+      for (int i = 0; i < 32; ++i) {
+        r = reduceOp(r, modifyOp(smem[i]));
+      }
```

I'll look at that performance.

dominikgrewe: I'd rather keep the `__syncthreads`. I doubt that removing it will improve performance much and it seems safer to keep it. NVIDIA actually discourages you from relying on warp sizes in case it changes in future (http://docs.nvidia.com/cuda/kepler-tuning-guide/#warp-synchronous).

wickedfoo: Sounds good re: __syncthreads().

FYI, removing the __syncthreads() for warp synchronous programming (admittedly sketchy based on the wording above) takes timing from 129.85 us -> 85.214 us here.
Being safe again, using intra-warp shuffles (CC 3.0+) for the bottom level takes timing down to 85.214 us (threads write to smem as usual; first warp loads from smem with a warp-wide stride and then reduces via shuffles). Using intra-warp shuffles for two-level reduction (reduce within each warp using shuffles; lane 0 in each warp writes out value to SM; first warp loads those values and reduces within itself using shuffles) takes timing down to 75.870 us with a lot less smem, which will have occupancy benefits.

```
394.63ms  2.8160us              (1 1 1)       (512 1 1)        12  2.0480KB        0B         -           -   Tesla K40m (0)         1         7  reduceByWarp1(void) [352]
394.65ms  2.8800us              (4 1 1)       (512 1 1)        12  2.0480KB        0B         -           -   Tesla K40m (0)         1         7  reduceByWarp1(void) [354]
394.66ms  2.9440us             (16 1 1)       (512 1 1)        12  2.0480KB        0B         -           -   Tesla K40m (0)         1         7  reduceByWarp1(void) [356]
394.68ms  4.5440us             (64 1 1)       (512 1 1)        12  2.0480KB        0B         -           -   Tesla K40m (0)         1         7  reduceByWarp1(void) [358]
394.69ms  10.080us            (256 1 1)       (512 1 1)        12  2.0480KB        0B         -           -   Tesla K40m (0)         1         7  reduceByWarp1(void) [360]
394.70ms  34.015us           (1024 1 1)       (512 1 1)        12  2.0480KB        0B         -           -   Tesla K40m (0)         1         7  reduceByWarp1(void) [362]
394.74ms  129.85us           (4096 1 1)       (512 1 1)        12  2.0480KB        0B         -           -   Tesla K40m (0)         1         7  reduceByWarp1(void) [364]
394.87ms  2.9120us              (1 1 1)       (512 1 1)        12  2.0480KB        0B         -           -   Tesla K40m (0)         1         7  reduceByWarp2(void) [366]
394.87ms  2.7520us              (4 1 1)       (512 1 1)        12  2.0480KB        0B         -           -   Tesla K40m (0)         1         7  reduceByWarp2(void) [368]
394.88ms  2.7520us             (16 1 1)       (512 1 1)        12  2.0480KB        0B         -           -   Tesla K40m (0)         1         7  reduceByWarp2(void) [370]
394.89ms  4.2560us             (64 1 1)       (512 1 1)        12  2.0480KB        0B         -           -   Tesla K40m (0)         1         7  reduceByWarp2(void) [372]
394.89ms  9.5040us            (256 1 1)       (512 1 1)        12  2.0480KB        0B         -           -   Tesla K40m (0)         1         7  reduceByWarp2(void) [374]
394.91ms  28.543us           (1024 1 1)       (512 1 1)        12  2.0480KB        0B         -           -   Tesla K40m (0)         1         7  reduceByWarp2(void) [376]
394.95ms  85.214us           (4096 1 1)       (512 1 1)        12  2.0480KB        0B         -           -   Tesla K40m (0)         1         7  reduceByWarp2(void) [378]
395.04ms  2.4000us              (1 1 1)       (512 1 1)        10  2.0480KB        0B         -           -   Tesla K40m (0)         1         7  reduceByWarpWithShuffle1(void) [380]
395.04ms  2.5280us              (4 1 1)       (512 1 1)        10  2.0480KB        0B         -           -   Tesla K40m (0)         1         7  reduceByWarpWithShuffle1(void) [382]
395.05ms  2.5600us             (16 1 1)       (512 1 1)        10  2.0480KB        0B         -           -   Tesla K40m (0)         1         7  reduceByWarpWithShuffle1(void) [384]
395.05ms  3.8720us             (64 1 1)       (512 1 1)        10  2.0480KB        0B         -           -   Tesla K40m (0)         1         7  reduceByWarpWithShuffle1(void) [386]
395.06ms  8.5110us            (256 1 1)       (512 1 1)        10  2.0480KB        0B         -           -   Tesla K40m (0)         1         7  reduceByWarpWithShuffle1(void) [388]
395.07ms  25.568us           (1024 1 1)       (512 1 1)        10  2.0480KB        0B         -           -   Tesla K40m (0)         1         7  reduceByWarpWithShuffle1(void) [390]
395.10ms  79.134us           (4096 1 1)       (512 1 1)        10  2.0480KB        0B         -           -   Tesla K40m (0)         1         7  reduceByWarpWithShuffle1(void) [392]
395.18ms  1.6960us              (1 1 1)       (512 1 1)         8       64B        0B         -           -   Tesla K40m (0)         1         7  reduceByWarpWithShuffle2(void) [394]
395.18ms  1.9520us              (4 1 1)       (512 1 1)         8       64B        0B         -           -   Tesla K40m (0)         1         7  reduceByWarpWithShuffle2(void) [396]
395.19ms  1.9520us             (16 1 1)       (512 1 1)         8       64B        0B         -           -   Tesla K40m (0)         1         7  reduceByWarpWithShuffle2(void) [398]
395.19ms  2.7200us             (64 1 1)       (512 1 1)         8       64B        0B         -           -   Tesla K40m (0)         1         7  reduceByWarpWithShuffle2(void) [400]
395.20ms  5.9520us            (256 1 1)       (512 1 1)         8       64B        0B         -           -   Tesla K40m (0)         1         7  reduceByWarpWithShuffle2(void) [402]
395.21ms  19.904us           (1024 1 1)       (512 1 1)         8       64B        0B         -           -   Tesla K40m (0)         1         7  reduceByWarpWithShuffle2(void) [404]
395.23ms  75.870us           (4096 1 1)       (512 1 1)         8       64B        0B         -           -   Tesla K40m (0)         1         7  reduceByWarpWithShuffle2(void) [406]
395.31ms  3.1360us              (1 1 1)       (512 1 1)         8  2.0480KB        0B         -           -   Tesla K40m (0)         1         7  reduceTree(void) [408]
395.31ms  3.2000us              (4 1 1)       (512 1 1)         8  2.0480KB        0B         -           -   Tesla K40m (0)         1         7  reduceTree(void) [410]
395.32ms  3.4240us             (16 1 1)       (512 1 1)         8  2.0480KB        0B         -           -   Tesla K40m (0)         1         7  reduceTree(void) [412]
395.32ms  5.8240us             (64 1 1)       (512 1 1)         8  2.0480KB        0B         -           -   Tesla K40m (0)         1         7  reduceTree(void) [414]
395.33ms  15.744us            (256 1 1)       (512 1 1)         8  2.0480KB        0B         -           -   Tesla K40m (0)         1         7  reduceTree(void) [416]
395.35ms  55.998us           (1024 1 1)       (512 1 1)         8  2.0480KB        0B         -           -   Tesla K40m (0)         1         7  reduceTree(void) [418]
395.41ms  218.56us           (4096 1 1)       (512 1 1)         8  2.0480KB        0B         -           -   Tesla K40m (0)         1         7  reduceTree(void) [420]
```

lib/THC/THCTensorMath.cu
```diff
@@ -759,32 +627,33 @@ void THCudaTensor_cumprod(THCState *state, THCudaTensor *self, THCudaTensor *src
...
-*/
+   to preserve the location of contention (for example min/max operations)
+ */
 template<class BinaryFunction>
 __global__ void THCudaTensor_kernel_transformReduceOuterDimIndex(float *tgt1, float *tgt2,
                                                              float *src_,
-                                                             unsigned num_orows,
-                                                             unsigned num_irows,
-                                                             unsigned row_size,
+                                                             dim4 src_stride,
```
dominikgrewe: I think something went wrong when this was merged. This is the old version of these kernels that only support up to 4-dimensional tensors.

wickedfoo: yes, @soumith merged from an earlier point since our internal tree wasn't quite updated with latest. See my "merge issues here, since we maybe didn't have the latest? i didn't change this code." comments up above. He'll fix this.

lib/THC/THCTensorMath.cu
```diff
@@ -2270,3 +2189,127 @@ void THCudaTensor_indexSelect(THCState *state, THCudaTensor *res_, THCudaTensor
...
+  THCudaTensor_free(state, contigMask);
+  THCudaTensor_free(state, maskPrefixSum);
+}
+
+void THCudaTensor_maskedFillByte(THCState* state, THCudaTensor *tensor, THByteTensor *mask, float value)
+{
+  THLongStorage* maskSize = THByteTensor_newSizeOf(mask);
+  THCudaTensor* maskCuda = THCudaTensor_newWithSize(state, maskSize, NULL);
+  THLongStorage_free(maskSize);
+
```
dominikgrewe: You need to copy `mask` into `maskCuda`, otherwise it will be uninitialized.
Same in the two functions below.

test/test.lua
```diff
@@ -1250,6 +1250,47 @@ function test.get_device()
...
+        -- previous line guarantees synchronization with srcDevice
+        cutorch.withDevice(dstDevice, function() cutorch.synchronize() end)
+
+        local t2_max = t2:max()
+        assert(t2_max == 1, "bad copy, transposeSrc= " .. transposeSrc ..
+               " transposeDst= " .. transposeDst .. ". t2:max() = " .. t2_max)
+      end
+   end
+end
+
```
dominikgrewe: Can you also add some tests for the masked* functions?

dominikgrewe: And for sort?

lib/THC/THCApply.cuh
```diff
@@ -0,0 +1,694 @@
...
+#endif
+__global__ void
+THCudaTensor_pointwiseApply1(TensorInfo<IndexType> a,
+                             IndexType totalElements,
+                             Op op) {
+  const IndexType blockId = getLinearBlockId<IndexType>();
+  const IndexType startLinearIndex =
+    getStartLinearIndex<IndexType>(blockId);
+
+  if (blockId == getLastLinearBlockId<IndexType>()) {
```
wickedfoo: for massive tensors with >536 million elements, virtual (with stride 0) or physical, this logic is wrong, fixing...

dominikgrewe:
What's the status of this PR? Would be great to get this in.

soumith:
Tomorrow. Writing tests.

soumith:
added some tests and fixes for review. will squash the commits before merging.

test/test.lua
```diff
@@ -1268,6 +1309,179 @@ function test.reset_device()
    tester:assertTensorEq(tf, u:float(), 1e-6, "values not equal after restoring the RNG state")
 end
 
+function test.maskedSelect()
+   local n_row = math.random(minsize,maxsize)
+   local n_col = math.random(minsize,maxsize)
+
+   -- contiguous, no result tensor, cuda mask
+   local x = torch.randn(n_row, n_col):float()
+   local mask = torch.DoubleTensor():rand(n_row*n_col):mul(2):floor():byte():resize(n_row,n_col)
```
dominikgrewe: Nit: This can be shortened to
`torch.DoubleTensor(n_row, n_col):uniform():round():byte()`

test/test.lua
```diff
@@ -1268,6 +1309,179 @@ function test.reset_device()
...
+   -- indexing maskedFill a[a:gt(0.5)] for example
+   local x = gt:clone()
+   x:t()[x:t():gt(0.5)] = 334
+   local x_cuda = gt:cuda()
+   x_cuda:t()[x_cuda:t():gt(0.5)] = 334
+   tester:assertTensorEq(x, x_cuda:float(), 0.00001,
+			 "Error in maskedFill non-contiguous indexing x[x:gt(y)]")
+
+end
+
```
dominikgrewe: Can you add a test for sort?

soumith: yep, on it.

soumith:
Okay, TensorCopy was implemented according to the docs, but the docs were apparently wrong. So removed it. 

Wrote a simple unit test for sort to start with, but sort does not seem to be agreeing with CPU for some cases.

To not delay this PR any further, I'll factor Sort out of this PR and merge it tomorrow if everyone's okay: cc: @dominikgrewe

soumith:
okay, ready for squash & merge now. unit-tested and good to go:
- Faster tensor math (especially for non-contiguous cases)
- maskedSelect
- maskedFill
- TensorApply and TensorReduce kernels

test/test.lua
```diff
@@ -1250,6 +1268,47 @@ function test.get_device()
...
+        -- try to induce a race on t1
+        cutorch.withDevice(srcDevice, function() t1:fill(4) end)
+
+        -- only synchronize with dstDevice because
+        -- previous line guarantees synchronization with srcDevice
+        cutorch.withDevice(dstDevice, function() cutorch.synchronize() end)
+
+        local t2_max = t2:max()
+        assert(t2_max == 1, "bad copy, transposeSrc= " .. transposeSrc ..
```
dominikgrewe: Should this be `tester:assert`? There are a couple of other places that simply call `assert`.

dominikgrewe:
Just one small comment on the testing code, otherwise this looks good to me!

Thanks a lot for breaking up the monolithic THC.cu. That was overdue :)

soumith:
fixed the asserts and squashed. merged. thanks for reviewing.

Issue #119, opened 2015-03-19
Segmentation fault when cutorch.deviceReset() was called earlier


tkocisky:
The code below produces a segmentation fault when deviceReset is called. When it is not called, it seems to work.

a.lua
```lua
require 'cutorch'
 
cutorch.deviceReset()
cutorch.setDevice(1)
cutorch.manualSeed(1)
 
print('Here1')
local a = torch.CudaTensor(1,106):fill(1)
local b = torch.CudaTensor(106,2):fill(3)
local c = torch.CudaTensor(1,2):fill(7)
print('Here2')
c:addmm(a,b)
print('Here3')
 
print('Here1')
local a = torch.CudaTensor(1,10006):fill(1)
local b = torch.CudaTensor(10006,2):fill(3)
local c = torch.CudaTensor(1,2):fill(7)
print('Here2')
c:addmm(a,b)
print('Here3')
```

Output:
```
# CUDA_VISIBLE_DEVICES="1" th a.lua 
Here1   
Here2   
Here3   
Here1   
Here2   
Segmentation fault (core dumped)
```

Output without the deviceReset:
```
# CUDA_VISIBLE_DEVICES="1" th a.lua 
Here1   
Here2   
Here3   
Here1   
Here2   
Here3
```

tkocisky:
@dominikgrewe Thanks! The segfault seems to be fixed in the latest torch/cutorch, but I'm still getting an error in addmm when running this.

Could someone check whether the above code works for them without any errors?

dominikgrewe:
The code works for me. What error do you get?

tkocisky:
It might be my installation of Torch then. (Also I'm using cuda 7.0.28, 346.47, Ubuntu 14.04.)

```
# CUDA_VISIBLE_DEVICES="0" th b.lua 
Here1	
Here2	
/users/username/torch7/bin/luajit: b.lua:12: /tmp/luarocks_cutorch-scm-1-6419/cutorch/lib/THC/THCBlas.cu(261) : cublas runtime error : the GPU program failed to execute
stack traceback:
	[C]: in function 'addmm'
	b.lua:12: in main chunk
	[C]: in function 'dofile'
	/users/username/torch7/lib/luarocks/rocks/trepl/scm-1/bin/th:131: in main chunk
	[C]: at 0x004055b0
```

szagoruyko:
I have reported this here https://github.com/torch/cutorch/issues/124 already

tkocisky:
[...]

PR #121, opened 2015-03-27
Recreate cuBLAS handle on deviceReset.

dominikgrewe:
Only need to reset the cuBLAS handle for the current device, because
only resources associated with the current device will be reset by
cudaDeviceReset.


dominikgrewe:
Fixes #119

PR #123, opened 2015-04-01
adding sort + test, fixing max/min and sum tests to not fail occasionally

soumith:
- Adds sort (by @wickedfoo) + test
- Adds some small fixes to unit tests so that sum and max/min dont fail occasionally.
- compareFloatAndCuda now preserves non-contiguity of inputs on CUDA as well (so that tensors have same size/stride)


soumith:
the sort stuff is already reviewed by dominik.

PR #128, opened 2015-04-07
GPU multi-stream support in cutorch

soumith:
This implements a version of the CUDA stream/event API in cutorch.

Original Author: @wickedfoo

By default streams are not allocated, but reserveStreams can be used to create a number of streams that are available on every device. This API does not expose explicit stream or event creation/destruction due to complexity, but it gives all the functionality that one would need via allowing one to construct single- or multi-device one-way and two-way barriers at the lua level

###Functions:
- `cutorch.reserveStreams(n)`: creates n user streams for use on every device.
- `n = cutorch.getNumStreams()`: returns the number of user streams available on every device. By `default`, this is `0`, meaning only the default stream (stream 0) is available.
- `cutorch.setStream(n)`: specifies that the current stream active for the current device (or any other device) is `n`. This is preserved across device switches. `1-N` are user streams, `0` is the default stream.
- `n = cutorch.getStream()`: returns the current stream active. By default, returns `0`.
- `cutorch.setDefaultStream()`: an alias for `cutorch.setStream(0)`
- `cutorch.streamWaitFor(streamWaiting, {streamsToWaitOn...})`: A `1-to-N-way` barrier. `streamWaiting` will wait for the list of streams specified to finish executing all kernels/events/barriers. Does not block any of the streamsToWaitOn. Current device only.
- `cutorch.streamWaitForMultiDevice(deviceWaiting, streamWaiting, {[device]={streamsToWaitOn...}...})`: (`deviceWaiting`, `streamWaiting`) will wait on the list of (`device`, `streams`...) pairs; handles single or multiple device. `cutorch.streamWaitForMultiDevice, a, b, {[a]={streams...}})` is equivalent to `cutorch.setDevice(a); cutorch.streamWaitFor(b, {streams...})`.
- `cutorch.streamBarrier({streams...})`: an `N-to-N-way` barrier between all the streams; all streams will wait for the completion of all other streams on the current device only. More efficient than creating the same N-to-N-way dependency via `streamWaitFor`.
- `cutorch.streamBarrierMultiDevice({[device]={streamsToWaitOn...}...})`: As with streamBarrier but allows barriers between streams on arbitrary devices. Creates a cross-device N-to-N-way barrier between all (device, stream) values listed.
- `cutorch.streamSynchronize(stream)`: equivalent to `cudaStreamSynchronize(stream)` for the current device. Blocks the CPU until stream completes its queued kernels/events.

###Tests:
- Tests added to cutorch for streamWaitFor(), streamWaitForMultiDevice(), streamBarrier() and streamBarrierMultiDevice().
- Small fixes for deviceReset

###Additional:
- All cutorch has been converted to be adhering to this new streams API with the exception of thrust calls. That is also coming soon.
- cunn calls are already converted as well, will sync that soon.
- Multi-device checks have been added to ALL cutorch Tensor functions. If two tensors are on different devices, barring the copy operation, all other operations will assertively fail. cunn checks are also done and will be synced soon.

After cunn fixes go in as well, containers like nn.Concat and nn.Parallel can be made truly parallel-ly executable.
Fixes for the cudnn package are also ready and will be synced soon.

cc: @dominikgrewe @clementfarabet


soumith:
As an example usage, a quickly hacked up Concat module with each branch executing parallely in a different stream. 
https://gist.github.com/soumith/4b32590926773c7755b0

Not fully tested.

lib/THC/THCGeneral.c
```diff
@@ -17,22 +17,81 @@ void THCudaInit(THCState* state)
...
+
+  /* Start in the default stream on the current device */
+  state->currentPerDeviceStream = 0;
+  state->currentStream = NULL;
+}
+
+void THCudaShutdown(THCState* state)
+{
+  THCRandom_shutdown(state);
+  free(state->blasState);
```
dominikgrewe: We shouldn't free `state->blasState` before calling `THCudaBlas_shutdown`.

I know this code wasn't changed in this PR, but since it's being touched anyway, I thought we might as well fix it now.

lib/THC/THCGeneral.c
```diff
@@ -17,22 +17,81 @@ void THCudaInit(THCState* state)
...
+  free(state->rngState);
+  free(state->deviceProperties);
+  THCudaBlas_shutdown(state);
+
+  int deviceCount = 0;
+  int prevDev = -1;
+  THCudaCheck(cudaGetDevice(&prevDev));
+  THCudaCheck(cudaGetDeviceCount(&deviceCount));
```
dominikgrewe: The device count is already stored in `state->numDevices`, so there's no need to query it again.

lib/THC/THCGeneral.c
```diff
@@ -42,21 +101,153 @@ void THCudaInit(THCState* state)
...
+  if (state->currentStream !=
+      state->streamsPerDevice[device][state->currentPerDeviceStream]) {
+    THError("Unexpected stream state");
+  }
+
+  /* Reallocate all streams for the current device; the 0 stream
+     doesn't need updating */
+  for (int dev = 0; dev < state->numDevices; ++dev) {
+    for (int stream = 1; stream <= state->numUserStreams; ++stream) {
+      THCudaCheck(cudaStreamCreate(&state->streamsPerDevice[dev][stream]));
```
dominikgrewe: We should call `cudaStreamDestroy` before creating a new one, shouldn't we?

test/test.lua
```diff
@@ -34,6 +34,16 @@ local function isEqual(a, b, tolerance, ...)
    end
 end
 
+local function checkMultiDevice(x, fn, ...)
```
dominikgrewe: What's the purpose of this? Why are we running each function again on a different device?

lib/THC/THCGeneral.c
```diff
@@ -42,21 +101,153 @@ void THCudaInit(THCState* state)
...
+  }
+
+  state->currentStream = state->streamsPerDevice[device][stream];
+  state->currentPerDeviceStream = stream;
+  THCudaBlas_setStream(state, device, state->currentStream);
+}
+
+void THCState_setStreamForCurrentDevice(THCState *state, int stream)
+{
+  if (state->currentPerDeviceStream != stream)
```
dominikgrewe: I'm wondering if we should set the stream regardless of whether it's already set. We do the same for the device. It's necessary when the same Lua state is operated on in different threads. CUDA state is typically per-thread, so if you set the device on thread 1 and then launch kernels from device 2, it will just use the default device. I'm guessing `cublasSetStream` works the same way?

nicholas-leonard:
This is really sexy. Love the example.

lib/THC/THCGeneral.c
wickedfoo: No, cudaDeviceReset() destroys all state associated with that device; the device no longer knows about previous streams.

There were a couple of different bugs with the device reset in cutorch; one was that pointers to various states weren't updated, and the second is that p2p device access was also destroyed and needs to be restored.

It's a really dangerous call, I argue that device reset should be removed from cutorch entirely, especially since we have no way of resetting/cordoning off all memory allocated for that device previously.

e.g.,
```
void* p = NULL;
cudaMalloc(&p, 1000);
cudaDeviceReset();
cudaError_t err = cudaFree(p);
assert(err == cudaSuccess);
```

will fail, just as:

```
cudaStream_t stream;
cudaStreamCreate(&stream);
cudaDeviceReset();
cudaError_t err = cudaStreamDestroy(stream);
assert(err == cudaSuccess);
```

will fail.

test/test.lua
wickedfoo: to verify that the cross-device failures trigger

lib/THC/THCGeneral.c
dominikgrewe: Fair point. I didn't look at it in the context of deviceReset. Maybe we should add a comment to this and other methods saying that they assume that deviceReset has been previously called.

I'd be equally happy to just remove device reset from cutorch.

test/test.lua
dominikgrewe: I see. Didn't realize that the tensor is created on a different device than the one we run the kernel on.

lib/THC/THCGeneral.c
wickedfoo: cublasSetStream is relative to the cublasHandle_t, not the CPU thread. cudaSetDevice is thread-local state, so the other thread must call to set that when it comes into scope.

Crossing CPU threads with the same Lua state seems dangerous to me anyways, since there's no explicit memory barrier anywhere in Torch/cutorch; only whatever mutual exclusion device you are using must provide the memory barrier, but in some cases the compiler must see the memory barrier too in order to generate proper code, especially if you are within a function (e.g., write something out, expect some other thread to pick it up without a memory barrier in the way; there's no explicit guarantee in the C11/C++11 memory model that writes must be made visible to other threads without such a barrier.)

lib/THC/THCGeneral.c
wickedfoo: Let's remove it then, I'm all for that :)

soumith: in another PR, let's remove it.

wickedfoo:
And to be clear, `reserveStreams(n)` doesn't create `n` streams, it just ensures that >= `n` streams are currently available.

There's no way to privately reserve streams for usage, but it doesn't matter that much that you are scheduling kernels for execution on some thread when someone else is doing the same, since you bothensure all the dependencies you need; you'll just wind up with extra synchronization (as long as you don't create a deadlock situation; I wonder if you create a circular one-way wait what it does? hmm) There's an upper limit on the concurrent execution that you can get out of the GPU anyways.

The end game here is that the setDevice/setStream API is a low-level primitive that is really meant to be used by higher level systems for scheduling execution of a graph, not by day-to-day Torch users in their own code. It's a powerful gun that you can easily shoot yourself in the foot with :)

lib/THC/THCGeneral.c
soumith: will change.

lib/THC/THCGeneral.c
soumith: will fix.

soumith:
made changes requested by dominik. merged.

soumith:
@nicholas-leonard to be clear, this wont work as you expect until we sync the cunn changes as well, because kernel launches have to take the stream argument.

wickedfoo:
(and fyi there seems to be a problem with cudnn, in that it is running kernels on the proper stream, but they are being serialized in execution with respect to each other. that's something we're going to take up with nvidia)

szagoruyko:
do you have plans to make backward running updateGradInput and accGradParameters in different streams?

dominikgrewe:
I have a question about the multi-device checks:
The whole point of peer access is that GPU1 can directly access GPU2's memory when performing an operation. So why are we disallowing it now?

wickedfoo:
Cross-device access is an order of magnitude (10x+) slower, because it is limited to PCIe or whatever bandwidth, and only upon page faults (it is faster to copy data to where you want to operate on it, than to operate on it remotely, on demand). Furthermore, if you do read data cross-device, you have to ensure that there are no read-after-write races and the like, which may be non-trivial to ensure.

It has been a significant issue for us when parallelizing code across GPUs that, for some lack of proper data placement, various kernels are reading data across devices resulting in significant slowdown, so we decided that it was a good idea to just disallow it, leaving copy as the only primitive that allows for cross-device access.

dominikgrewe:
You could just disable peer access to achieve the same, couldn't you? Copies also work between GPUs that are not peers.

soumith:
@dominikgrewe I think the idea is to leave the option open, but keep checks in-place to not shoot oneself in the foot.

Stuff like our custom non-contiguous copy kernel will not be possible if we disable p2p for example.

dominikgrewe:
Then maybe we should have an easy way to disable the checks?

With the non-contiguous copy kernels relying on p2p, they currently fail when copying data between GPUs that don't have p2p access. Should we fall back to contiguous copies between devices in those cases?

wickedfoo:
Copies (cudaMemcpyAsync) are slower between GPUs when you disable p2p access; they have to be intermediated by the host. If p2p access is available, I think it should always be enabled.

dominikgrewe:
Agreed, we should always enable p2p access if available. But we should also make sure that our code works if we don't have p2p access. I'll file a bug for the non-contiguous copy.

wickedfoo:
Sounds good.

koraykv:
I just want to make sure that one can easily turn off p2p checks. I don't think p2p operations are necessarily slower, I think it might depend on the hardware since we have benchmarks showing speedups actually. I suggest we put a compile time flag to disable these checks.

soumith:
@koraykv compile-time flag sounds fine to me. Even runtime flag is fine, doesn't really eat performance. I'll prepare patches for this. I'd be interested to know when p2p operations are actually faster, because they (1) introduce a sync point (2) they copy over memory to the other GPU page-by-page, do the operation, and throw it away, so many small copies over and over again interleaved with computation in a synchronous fashion.

dominikgrewe:
On simple benchmarks it is faster (see below for example). As you said, it introduces a sync point, which may or may not be a problem. But we shouldn't disallow it completely.

```lua
require 'cutorch'

local size = math.pow(2, 28)

cutorch.setDevice(1)
local x1 = torch.CudaTensor(size):uniform()
local tmp1 = torch.CudaTensor(size)
cutorch.synchronize()

cutorch.setDevice(2)
local x2 = torch.CudaTensor(size):uniform()
cutorch.synchronize()

cutorch.setDevice(1)
local start = torch.tic()
x1:add(x2)
cutorch.synchronize()
local time = torch.toc(start)
print('direct: ' .. time)


start = torch.tic()
tmp1:copy(x2)
x1:add(tmp1)
cutorch.synchronize()
local time = torch.toc(start)
print('indirect: ' .. time)
```

PR #133, opened 2015-04-10
Implementation of logical all and any.

dominikgrewe:


dominikgrewe:
Are there any plans to have reduction kernels across all dimensions similar to the ones for reductions across a specific dimension?

soumith:
We are atm focusing on multi-GPU in our current sprint, so no plans from us in the very near future.

wickedfoo:
@dominikgrewe I will add this on my lengthy to do list. I want to do it sometime :)

A facility which I would want to have for doing this are device memory scratchpads in Torch. If the reduction is large enough, Thrust first reduces within blocks into a temporary scratch space (say, one float per CTA block), and then launches a second kernel to reduce across all blocks to a single value. If the reduction is small enough, there is enough parallelism in one CTA for doing this, but many cases require more than one CTA.

In order to avoid a cudaMalloc/Free pair, I would implement a permanently resident reduction scratch space on every device equal to # of streams allocated * # of SMs * 4-8 * sizeof(float), or 240-480 bytes per stream that you have allocated (it's per each stream because kernels launched could execute concurrently, and the '4-8' is to allow a tradeoff between more threads/block with fewer blocks vs fewer threads/block with more blocks depending upon smem usage. Really the occupancy calculator should be used here based on smem/register count, but tbd). The addresses would be managed by the THCState.

There's still a sync issue because Torch doesn't have the concept of a lazily evaluated scalar value (e.g., a CudaTensor(1)), so it will synchronize when copying the value back to the CPU for use there. Plus, that scalar value would probably have to come from this scratch space (or would come from a usual cudaMalloc/Free), which gets invalidated at the next reduction kernel launch, unless we want to allocate for each call.

PR #136, opened 2015-04-18
Auto device mode, plus allocation helper functions.

adamlerer:
This diff introduces an alternative way of writing multi-GPU cutorch
code. In this mode, the location of each tensor is specified, and the
appropriate GPU for each kernel is determined automatically based on the
location of its argument tensors. It's backwards-compatible and interoperable
with the old-style multi-GPU API.

-------------------------------------------------------

Calling cutorch.setDevice(0) causes cutorch to automatically use the appropriate device for a given computation. Non-empty tensors cannot be allocated in this mode, unless one of the following two convenience methods is used:

    torch.CudaDeviceOn(device, ...)
    torch.FloatTensor(...)):cuda(device)

Example:

    > cutorch.setDevice(0)  -- 'auto' or 'no device'
    > a = CudaTensorOn(1, 10)  -- size 10, on device 1
    > b = CudaTensorOn(2, 10)  -- size 10, on device 2
    > c = CudaTensor() -- size 0, unallocated
    > c:add(a, 1)   -- c is now size 1, on device 1
    > d = a + b
    Error: tensors are on different devices

The old behavior w.r.t empty tensors is preserved - they are unallocated, and are assigned a device only when resized to non-zero size. However, you can now assign an empty tensor to a device, using either t:setDevice(device) or one of the convenience methods above, and in device=0 mode it will retain its assigned device upon resize.

@dominikgrewe can you take a look? @soumith suggested I get your feedback. Thanks!


dominikgrewe:
Thanks Adam.

I haven't looked at the code closely yet, but I noticed one thing:
I think it may be too late to set the GPU during `checkGPU`. For example, if you call `c = torch.mm(a, b)`, `c` will be allocated on the current device, which isn't necessarily the same device that `a` and `b` are on.
Any ideas how that could be fixed?

adamlerer:
Hi Dominik,

Yes, that's really the crux of this feature. In fact, `c` will be allocated on the correct device.

    > require 'torch'
    > require 'cutorch'
    > cutorch.setDevice(0)
    > a = torch.CudaTensorOn(2,2,2):zero()
    > b = torch.CudaTensorOn(2,2,2):zero()
    > c = torch.mm(a, b)
    > print( c:getDevice() )
    2

The reason this works is that empty tensors are not assigned a device. torch.mm allocates an empty thensor for _res, then checkGPU is called (which switches to device 2), and then _res is assigned, at which point it gets assigned to device 2.

dominikgrewe:
This isn't quite true. It only happens to work in this case because the last device used is the one that a and b were allocated on. If you add `x = torch.CudaTensorOn(1, 2)` before the call to `torch.mm` it won't work any more. The reason is that `torch.mm` actually resizes the tensor before calling into C (and thus calling checkGPU).

adamlerer:
Oh yes, you're right. Fixing these particular cases (looks like there are four of them within cwrap) wasn't difficult. For example, I can just add a checkGPU line within the cwrap init function. You are correct, however, that auto device handling only works to the extent that it's wrapped in checkGPU calls. Library functions that want to support auto device mode will have to do so explicitly.

dominikgrewe:
I'm a bit worried that there'll be more cases like this and we need more hacks to work around them. I'm not convinced the convenience we gain outweighs the additional complexity. The original model was very straightforward: you set the device, and everything happens on that device. In auto mode this isn't as obvious any more. What happens when we call `cutorch.synchronize()`, for example? We could pass in a device ID, of course, but there'll probably be more cases like this that we haven't thought about.

@soumith What do you think? I'm not entirely sold yet, but I'm happy to be convinced otherwise :)

soumith:
In general, I am FOR an auto-mode, but I think choosing semantics for it should be done very carefully.

The most compelling reason for having an auto-mode is an example of an nngraph or nn model that spans multiple-GPUs (and does not fit in the simplistic DataParallel or ModelParallel paradigm), or a torch script that does lots of computation across GPUs.
I've asked Adam to minimize one of our internal scripts and provide it as an example.
The amount of sprinkling calls to cutorch.WithDevice makes using multiple GPUs embarrassingly ugly.

The current iteration of the semantics of auto-mode were carefully debated and chosen, but there are definitely corner cases like cutorch.synchronize that we did not think about.

However, after looking at Adam's example (that he will provide), if you think of other ideas that would make programming for multi-GPU in a single script somewhat reasonably non-ugly, I'd love to hear some ideas.

adamlerer:
@dominikgrewe here's an example of a simple multi-GPU model with and without auto device mode.

Manual:

    net = nn.Sequential()
    for i = 1,3 do
      cutorch.setDevice(i)
      net:add(nn.Linear(1000,1000):cuda())
      net:add(nn.SoftMax():cuda())
    end

    function fp(input)
      for i = 1,3 do
        cutorch.setDevice(i)
        input = net.modules[2*i-1]:forward(input)
        input = net.modules[2*i]:forward(input)
        cutorch.setDevice(i+1)
        local tmp = torch.CudaTensor(input:size()):copy(input)
        input = tmp
      end
      return input
    end

    function bp(input, gradOutput)
      -- a similar mess here
    end

    cutorch.setDevice(1)
    local input = torch.CudaTensor(1000)
    local output = fp(input)
    print(output:getDevice())

    local gradOutput = output/100
    local gradInput = bp(input, gradOutput)
    print(gradInput:getDevice())

Auto (requires a few small changes to nn/cunn that I haven't pushed yet):

    cutorch.setDevice(0)

    net = nn.Sequential()
    for i = 1,3 do
      net:add(nn.Linear(1000,1000):cuda(i))
      net:add(nn.SoftMax():cuda(i))
      net:add(nn.TransferDevice(i, i+1)) -- a new little shim in cunn
    end

    local input = torch.CudaTensorOn(1, 1000)
    local output = net:forward(input)
    print(output:getDevice())

    local gradOutput = output/100
    local gradInput = net:backward(input, gradOutput)
    print(gradInput:getDevice())

This is a toy example, but I think you can imagine what a nightmare the manual approach becomes for larger modules with more complex and dynamic parallelization. With the manual approach there's no separation of concerns between specifying the computation and the parallelization.

You make a good point about cutorch.synchronize() - I need to go back and make sure this and the other cutorch-global functions have reasonable semantics in auto-device mode. I do believe, though, that these semantics, which make explicit the device rather than having the device specified by global (thread-local) state, will make the API less error-prone.

dominikgrewe:
If each module knows which device it's being run on, why can't the module itself call `setDevice`? All you'd need to do is wrap the `forward` and `backward` functions (this can be done automatically when calling `cuda(i)`) and you'd get the same behaviour afaict. Or are there cases where this wouldn't work?

I like things like `CudaTensorOn` and `cuda(deviceId)`. All I'm worried about is the complexity of supporting different modes in cutorch, where we constantly need to think about what will happen in each case.

dominikgrewe:
In any case, if we're going ahead with this, we need to make sure that the semantics are clear and documented. As you said, Adam, there are a couple of cutorch-global functions that we'd need to find intuitive behaviour for.

adamlerer:
Yeah, after I wrote out the example yesterday, @soumith and I also started talking about whether we should just wrap nn modules to set the device, instead of the more heavyweight changes to cutorch.  But I'd prefer something that worked for cutorch as well, because the problems with setDevice() are not restricted to cunn layers. In the scripts that our users are hacking on, they're using both cutorch calls and cunn layers across multiple devices.  

I understand that what you're objecting to here is the introduction of another global mode, which adds its own new semantics and hidden state that needs to be understood every time someone wants to add something new to cutorch. I agree with that - originally, I just wanted `checkGPU` to always set the correct device instead of asserting, and restore it afterwards. Unfortunately, restoring it is a bit tricky, because you have to handle all possible control flows (C functions calling other C functions, different exit points, THError, ...), and if I don't restore the device, you can get unexpected behavior (e.g. stuff within a withDevice() call not using the correct device). The idea behind `device=0` mode was that I wouldn't have to restore the device.

I think we agree that `CudaTensorOn()` and `cuda(i)` are useful (and thus tensor:setDevice() and devices for empty tensors are useful as well). What other alternatives are there for cutorch math?

1) Wrap cuda math functions within lua to run with the correct device. This is what some of our users are actually doing right now in a very hacky way to work around the device hell - something like

    for k, v in _cudaTensorMeta do
        if type(v) == 'function" and EXCLUDED[k] == nil then
           _cudaTensorMeta[k] = <wrapper that sets and restores device>
        end
    end
    <same for cutorch>

This approach has some advantages - you don't have an extra mode that global functions like synchronize() need to handle appropriately. However, you still have to carefully consider the semantics of "exceptional" functions like __gc, getDevice, resizeAs, etc. where you don't want to set the device, and you can get some really nasty bugs if you do it wrong. It also may adversely affect debugging and performance.

2) I could just add some shorthand that would make it easier to control the device manually.

    function CudaTensor:onDevice(f)
        cutorch.setDevice(self:getDevice(), f)
        return self
    end

    a = torch.mm(b:onDevice(), c)
    d:onDevice():add(1)
 
You'd just have to be careful to call onDevice() within every call. I have a feeling that users would be annoyed by this.

Getting back to your earlier concerns about my approach's interaction with cutorch.synchronize() etc., I feel that these functions are infrequently used, and I'm inclined to just assert against calling the device-specific ones in no-device mode, with an error message "no device specified. Use cutorch.setDevice() or cutorch.withDevice() to specify a device".

dominikgrewe:
I agree. If you want full support for this in cutorch, you probably won't get around to adding support right into THC. Even if you could wrap all math operations, you'd still have the issue of CUDA code in nn modules.

About functions like cutorch.synchronize: How about we give all of them an optional deviceId parameter (if applicable). That may be useful for manual mode too. It would also be cleaner because currently some functions take a device ID while others don't. These functions can then just error out when no device is specified in auto mode, as you suggested.

adamlerer:
@dominikgrewe I added a device parameter to all the device-specific cutorch functions. Let me know if I missed any. Thanks!

Storage.c
```diff
@@ -140,4 +153,15 @@ void cutorch_CudaStorage_init(lua_State* L)
       lua_pop(L, 1);
     }
   }
+
+  luaT_pushmetatable(L, "torch.CudaStorage");
+  lua_pushcfunction(L, cutorch_CudaStorage_getDevice);
+  lua_setfield(L, -2, "getDevice");
+  lua_pop(L, 1);
+
+  luaT_pushmetatable(L, "torch.CudaStorage");
```
dominikgrewe: No need to pop the meta table just to immediately push it again.

TensorMath.lua
```diff
@@ -96,6 +96,8 @@ wrap.types.CudaTensor = {
       local txt = {}
       if arg.creturned then
          -- this next line is actually debatable
+         -- doesn't this guarantee a leak in some cases?
+         -- I suggest we assert against creturned tensors...
```
dominikgrewe: Good point. I guess it's not really clear whether the function that returns the tensor keeps a reference to it or not. Seems like this isn't used at the moment, so maybe we should just assert against it, as you suggested.

dominikgrewe:
cutorch_getDeviceProperties always expects a device at the moment. We should make that optional as in all the other functions.

init.c
```diff
@@ -184,9 +184,31 @@ void waitMultiDeviceEvents(lua_State *L, THCState *state, int arg,
...
+/*
+   Usage:
+   cutorch.synchronize([device])
+   Synchronizes the current or specified device with the host.
+*/
 static int cutorch_synchronize(lua_State *L)
 {
+  int oldDevice = optSetDevice(L, 1, "synchronize");
   THCudaCheck(cudaDeviceSynchronize());
+  THCState_setDevice(cutorch_getstate(L), oldDevice);
```
dominikgrewe: Is it worth passing a closure into optSetDevice (similar to withDevice)? Then you wouldn't have to reset the device in every single function using it.

lib/THC/THCTensor.c
```diff
@@ -599,7 +599,7 @@ void THCudaTensor_freeCopyTo(THCState *state, THCudaTensor *self, THCudaTensor *
 static void THCudaTensor_rawInit(THCState *state, THCudaTensor *self)
 {
   self->refcount = 1;
-  self->storage = NULL;
+  self->storage = THCudaStorage_new(state);
```
dominikgrewe: I'm a bit worried about changing the behaviour here. THCudaTensor_data, for example, behaves slightly different now for empty tensors. Previously it would return `nil`, now it returns a pointer value `NULL` (causing different behaviour in conditionals, for example). I haven't seen any other cases of changing behaviour, but that doesn't mean there aren't any.
Could a tensor also store a device ID? Then you wouldn't have to create an empty storage, right?

lib/THC/THCTensor.c
```diff
@@ -758,19 +759,35 @@ float THCudaTensor_get4d(THCState *state, const THCudaTensor *tensor, long x0, l
...
-  int valid = 1;
   for (unsigned int i = 0; i < nTensors; i++) {
     THCudaTensor* tensor = va_arg(args, THCudaTensor*);
     int tensorDev = THCudaTensor_getDevice(state, tensor);
-    if (tensorDev != -1 && tensorDev != curDev) {
-      valid = 0;
-      break;
+    if (tensorDev != THC_DEVICE_NONE) {
+      if (kernelDev != tensorDev && kernelDev != THC_DEVICE_NONE) {
+        return 0; // device mismatch
```
dominikgrewe: Make sure `va_end` gets called before the function returns.

init.c
adamlerer: It's a matter of personal preference; IMO the way it's written is more clear. I'm open to changing it if you feel strongly.

lib/THC/THCTensor.c
adamlerer: Sorry, you lost me here. Reading the code, it looks like it will return `0` (aka `NULL`) in both cases. I'm not sure what it means for a C function to return `nil`.

    float *THCudaTensor_data(THCState *state, const THCudaTensor *self)
    {
      if(self->storage)
        return (self->storage->data+self->storageOffset);
      else
        return NULL;
    }

lib/THC/THCTensor.c
adamlerer: Ouch; good catch.

lib/THC/THCTensor.c
adamlerer: To address your higher-level point, I think that a device really should be assigned to a storage, because multiple tensors may be assigned to the same storage and should therefore have the same device automatically without trying to maintain that invariant manually.

However, the semantics have shifted a bit since I did this implementation - perhaps an unallocated tensor should have no storage, but once I call t:setDevice(i), it gets an empty storage with a device. Maybe then I wouldn't need this THC_DEVICE_NONE indicator, because every storage would have a real device.

Tensor.c
```diff
@@ -224,6 +231,10 @@ void cutorch_CudaTensor_init(lua_State* L)
   luaT_pushmetatable(L, "torch.CudaTensor");
   lua_pushcfunction(L, cutorch_CudaTensor_getDevice);
   lua_setfield(L, -2, "getDevice");
+  lua_pop(L, 1);
 
+  luaT_pushmetatable(L, "torch.CudaTensor");
```
dominikgrewe: Also remove the pop & push here.

init.c
dominikgrewe: I feel it would be less error-prone. You can't accidentally forget to reset the device. If you call it something like `withOptDevice` it should be fairly clear what's going on.

lib/THC/THCTensor.c
dominikgrewe: That sounds good to me. I'm just a bit worried to diverge from Torch's behaviour, but maybe it's not a problem

Re the return value of THCudaTensor_data, I'm not sure myself why it seems to behave differently. But without your patch calling `torch.CudaTensor():data` from Lua returns `nil`, whereas with your patch it returns `cdata<float *>: NULL`. Those two values behave differently in conditionals, so if you have `if x:data() then ...` in your code, the behaviour will now change for empty tensors.

dominikgrewe:
Thanks Adam. Could you change cutorch_getDeviceProperties as I mentioned in an earlier comment?
Tensor cloning also doesn't seem to work in auto mode. The new tensor isn't necessarily on the device as the original tensor, as it should be intuitively.

init.c
dominikgrewe: Actually, ignore that :) I thought we were in Lua, where doing this would be easy, but in C it would be a bit of a pain...

adamlerer:
Hi Dominik, sorry for the delay, I was out of town. `clone` is actually a tricky case - I'm not sure if it's better to just make it an error, because

    b = a:clone()

should be equivalent to

    b = CudaTensor()
    b:resizeAs(a):clone(a)

which is an error (in `resizeAs`) when `device=0`.

I assume that other people's intuition will match yours though, and made `clone` behave as you described.

dominikgrewe:
Thanks. If you could remove the pop & push in `cutorch_CudaTensor_init` and squash your commits, I think this is good to go.

adamlerer:
Done! I will submit a PR to cunn next with auto-allocation tweaks.

dominikgrewe:
Thanks Adam!

adamlerer:
Thanks Dominik! 

I actually decided to replace `cuda(device)` with `cudaOn(device)` to avoid overloading ambiguity. I'll send you a PR for this (as well as PR's for nn and cunn) later today.

soumith:
@adamlerer can you add to README.md the semantics of the auto-mode. Most of the comments and summary in this PR reformatted should be sufficient.

adamlerer:
@soumith The cutorch README.md is pretty empty - there are just a few bullet points about streams. I can add some comments about devices, but it really needs to be written in full.

soumith:
@adamlerer added a slightly more complete README.md just now.

dominikgrewe:
Great, thanks a lot Adam! Are you going to add documentation on auto-mode?

PR #138, opened 2015-04-20
Fix squeeze for 1-element tensors.

dominikgrewe:


soumith:
Thanks. I'll patch squeeze1d (that's right above it) as well. I patched it internally but didn't push the fix on friday.

dominikgrewe:
Thanks. Overlooked squeeze1d, sorry.

PR #141, opened 2015-04-29
Tensor.lua: return `self` on `apply`

deltheil:
To comply with non-CUDA `apply` version (see [here](https://github.com/torch/torch7/blob/ee03a0a/doc/tensor.md#self-applyfunction)).


dominikgrewe:
Thanks!

PR #144, opened 2015-04-30
Auto device: API changes, bug fixes, README.md

adamlerer:
- Change :cuda(device) overload to :cudaOn(device)
- Add :cloneOn(device)
- Fix bug in +,-,*,/ metamethods: checkGPU wasn't being called on these
  metamethods.
- Add description of auto-device mode to README.md
- cutorch.test(): add seed argument and print seed, so that failures can be reproduced

### DIFF
--- a/torch/generic/Tensor.c
+++ b/torch/generic/Tensor.c
@@ -240,6 +240,22 @@ static int torch_Tensor_(clone)(lua_State *L)
   return 1;
 }
 
+static int torch_Tensor_(cloneOn)(lua_State *L)
+{
+  THCState *state = cutorch_getstate(L);
+  THTensor *self = luaT_checkudata(L, 1, torch_Tensor);
+  int device = luaL_checkint(L, 2)-1;
+  int oldDev = -1;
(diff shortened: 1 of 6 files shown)


soumith:
lgtm, will let @dominikgrewe merge.

torch/generic/Tensor.c
```diff
@@ -240,6 +240,22 @@ static int torch_Tensor_(clone)(lua_State *L)
   return 1;
 }
 
+static int torch_Tensor_(cloneOn)(lua_State *L)
```
dominikgrewe: Do we still need the logic in `torch_Tensor_(clone)` (above) for auto-device mode? Maybe we should just assert that we're in manual mode (same for `cuda`), which would be more in line with the documentation.

adamlerer: The one reason I like the given behavior is that there might be clone()
calls buried inside modules, that I still want to work in auto-mode.

dominikgrewe: Fair enough. There's no harm in leaving it there.

PR #150, opened 2015-05-07
Include THAtomic.h

dominikgrewe:
Follow-up from #149 
Didn't see the warnings earlier, sorry.


soumith:
Is thought it gets included transitive ly through TH.h

dominikgrewe:
We only include TH.h in THCGeneral.c

PR #158, opened 2015-05-22
Adding support for Storage views to cutorch.

zakattacktwitter:
This is the same as https://github.com/torch/torch7/pull/240 but for cutorch.


dominikgrewe:
I think this broke the build. You need to add `view` to THCudaStorage (https://github.com/torch/cutorch/blob/master/lib/THC/THCStorage.h#L12)

dominikgrewe:
And THCStorage_retain gets `state` as a first argument. I'm preparing a patch.

zakattacktwitter:
Thanks.

PR #159, opened 2015-05-22
Fixing call to cudaMemGetInfo to use the correct device.

zakattacktwitter:
The function cudaMemGetInfo works on the currently set device, but this function allows an arbitrary device to be passed in. See:

http://developer.download.nvidia.com/compute/cuda/4_1/rel/toolkit/docs/online/group__CUDART__MEMORY_gd5d6772f4b2f3355078ecd6059e6aa74.html


dominikgrewe:
Good catch. Thanks!

PR #160, opened 2015-05-26
Added compile-time flag to disable GPU checks

yozw:


soumith:
thanks, I was supposed to do this, but it was lost in my list of things.

yozw:
No problem :)

PR #162, opened 2015-05-27
Small fix to cutorch install.

zakattacktwitter:
Install and build are two distinct things.

### DIFF
--- a/rocks/cutorch-scm-1.rockspec
+++ b/rocks/cutorch-scm-1.rockspec
@@ -20,7 +20,7 @@ dependencies = {
 build = {
    type = "command",
    build_command = [[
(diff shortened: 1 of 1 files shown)


dominikgrewe:
This seems to slow down my install quite a bit. Seems like `make install` actually builds a lot of files (again?). I'm not very familiar with the build scripts, so not sure what's going on.

soumith:
yea, I made the earlier change precisely because nvcc/cmake are recompiling the files twice. Once on "make" and the second time on "make install".

zakattacktwitter:
Thanks guys, just so you know I am going through 6 months of changes to our internal fork of torch source and picking out bits to merge back into the public repo, before i re-sync us with the HEAD of the public repos. Almost all of these changes were not made by me, its a lot of forensics. Thanks for your patience as I sort through it all.

soumith:
@zakattacktwitter I'm really happy that you guys are syncing back up with HEAD. We will be super patient I promise. I went through the same process 6 months ago, when I just joined FB.

PR #167, opened 2015-05-28
Streams for BLAS. maskedCopy. generic reduce kernels. bugfix.

soumith:
Stream support for BLAS Handles.
maskedCopy implemented
generic Reduce kernels
Fixed a small ffi inconsistency introduced with #158

### DIFF
--- a/FFI.lua
+++ b/FFI.lua
@@ -4,16 +4,28 @@ if ok then
    local cdefs = [[
 typedef struct CUstream_st *cudaStream_t;
 
+struct cublasContext;
+typedef struct cublasContext *cublasHandle_t;
+typedef struct CUhandle_st *cublasHandle_t;
+
+typedef struct _THCCudaResourcesPerDevice {
+  cudaStream_t* streams;
+  cublasHandle_t* blasHandles;
+} THCCudaResourcesPerDevice;
+
+
 typedef struct THCState
 {
   struct THCRNGState* rngState;
-  struct THCBlasState* blasState;
   struct cudaDeviceProp* deviceProperties;
   cudaStream_t currentStream;
-  cudaStream_t** streamsPerDevice;
+  cublasHandle_t currentBlasHandle;
+  THCCudaResourcesPerDevice* resourcesPerDevice;
   int numDevices;
   int numUserStreams;
+  int numUserBlasHandles;
   int currentPerDeviceStream;
+  int currentPerDeviceBlasHandle;
   struct THAllocator* cudaHostAllocator;
 } THCState;
 
@@ -27,6 +39,7 @@ typedef struct THCudaStorage
     char flag;
     THAllocator *allocator;
     void *allocatorContext;
+    struct THCudaStorage *view;
 } THCudaStorage;
 
 typedef struct THCudaTensor
--- a/lib/THC/THCGeneral.c
+++ b/lib/THC/THCGeneral.c
@@ -4,6 +4,12 @@
 #include "THCBlas.h"
 #include "THCAllocator.h"
 
+/* Size of scratch space available in global memory per each SM + stream */
+#define GLOBAL_SCRATCH_SPACE_PER_SM_STREAM 4 * sizeof(float)
+
+THCCudaResourcesPerDevice* THCState_getDeviceResourcePtr(
+  THCState *state, int device);
+
 void THCudaInit(THCState* state)
 {
   int count = 0;
@@ -15,9 +21,6 @@ void THCudaInit(THCState* state)
   state->rngState = (THCRNGState*)malloc(sizeof(THCRNGState));
   THCRandom_init(state, count, device);
 
-  state->blasState = (THCBlasState*)malloc(sizeof(THCBlasState));
-  THCudaBlas_init(state, count, device);
-
   THCAllocator_init(state);
 
   state->numDevices = count;
@@ -25,21 +28,33 @@ void THCudaInit(THCState* state)
     (struct cudaDeviceProp*)malloc(count * sizeof(struct cudaDeviceProp));
 
   state->numUserStreams = 0;
-  state->streamsPerDevice =
-    (cudaStream_t**)malloc(count * sizeof(cudaStream_t*));
+  state->numUserBlasHandles = 0;
 
   /* Enable P2P access between all pairs, if possible */
   THCudaEnablePeerToPeerAccess(state);
 
-  for (int i = 0; i < count; ++i)
-  {
+  state->resourcesPerDevice = (THCCudaResourcesPerDevice*)
+    malloc(count * sizeof(THCCudaResourcesPerDevice));
+  for (int i = 0; i < count; ++i) {
+    THCCudaResourcesPerDevice* res = THCState_getDeviceResourcePtr(state, i);
+
     THCudaCheck(cudaSetDevice(i));
     THCudaCheck(cudaGetDeviceProperties(&state->deviceProperties[i], i));
     /* Stream index 0 will be the default stream for convenience; by
        default no user streams are reserved */
-    state->streamsPerDevice[i] =
-      (cudaStream_t*)malloc(sizeof(cudaStream_t));
-    state->streamsPerDevice[i][0] = NULL;
+    res->streams = NULL;
+    res->blasHandles = NULL;
+
+    /* The scratch space that we want to have available per each device is
+       based on the number of SMs available per device */
+    int numSM = state->deviceProperties[i].multiProcessorCount;
+    size_t sizePerStream = numSM * GLOBAL_SCRATCH_SPACE_PER_SM_STREAM;
+    res->scratchSpacePerStream = sizePerStream;
+
+    /* Allocate scratch space for each stream */
+    res->devScratchSpacePerStream = malloc(sizeof(void*));
+    THCudaCheck(cudaMalloc(&res->devScratchSpacePerStream[0],
+                           sizePerStream));
   }
 
   /* Restore to previous device */
@@ -48,32 +63,52 @@ void THCudaInit(THCState* state)
   /* Start in the default stream on the current device */
   state->currentPerDeviceStream = 0;
   state->currentStream = NULL;
+
+  /* There is no such thing as a default cublas handle.
+     To maintain consistency with streams API, handle 0 is always NULL and we
+     start counting at 1
+   */
+  THCState_reserveBlasHandles(state, 1);
+  state->currentPerDeviceBlasHandle = 1;
+  state->currentBlasHandle = THCState_getDeviceBlasHandle(state, device, 1);
 }
 
 void THCudaShutdown(THCState* state)
 {
   THCRandom_shutdown(state);
-  THCudaBlas_shutdown(state);
   THCAllocator_shutdown(state);
-  free(state->blasState);
   free(state->rngState);
   free(state->deviceProperties);
 
+  int deviceCount = 0;
   int prevDev = -1;
   THCudaCheck(cudaGetDevice(&prevDev));
+  THCudaCheck(cudaGetDeviceCount(&deviceCount));
 
-  for (int dev = 0; dev < state->numDevices; ++dev) {
+  for (int dev = 0; dev < deviceCount; ++dev) {
     THCudaCheck(cudaSetDevice(dev));
-
     /* Free Torch-defined streams (0 is the default stream) */
     for (int stream = 1; stream <= state->numUserStreams; ++stream) {
-      THCudaCheck(cudaStreamDestroy(state->streamsPerDevice[dev][stream]));
+      THCudaCheck(cudaStreamDestroy(
+                    THCState_getDeviceStream(state, dev, stream)));
+    }
+    /* Free Torch-defined handles (0 is NULL for consistency with streams API) */
+    for (int handle = 1; handle <= state->numUserBlasHandles; ++handle) {
+      THCudaCheck(cublasDestroy(
+                    THCState_getDeviceBlasHandle(state, dev, handle)));
+    }
+    /* Free per-stream scratch space; starts at 0 because there is space for
+       the default stream as well*/
+    for (int stream = 0; stream <= state->numUserStreams; ++stream) {
+      THCudaCheck(cudaFree(THCState_getDeviceScratchSpace(state, dev, stream)));
     }
 
-    free(state->streamsPerDevice[dev]);
+    free(state->resourcesPerDevice[dev].streams);
+    free(state->resourcesPerDevice[dev].blasHandles);
+    free(state->resourcesPerDevice[dev].devScratchSpacePerStream);
   }
+  free(state->resourcesPerDevice);
 
-  free(state->streamsPerDevice);
   THCudaCheck(cudaSetDevice(prevDev));
 }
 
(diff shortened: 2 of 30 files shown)


soumith:
all tests pass.

soumith:
cc: @dominikgrewe more stuff on it's way. do review when you get a chance.

We've been using this branch for most of the last month or more, so it's definitely stable and afaik fairly bug-free.

dominikgrewe:
Could we couple the BLAS handles with streams? I.e. for each stream, we'll have a separate BLAS handle and when switch streams you automatically switch BLAS handles.
What do you get from managing them separately?

dominikgrewe:
Btw, it would be great if you could break these patches down a bit more. There is so much going on here, it's hard to make sense of it all. Won't get a chance until early next week to take a closer look.

Was the THCDeviceTensor code meant to be in this PR? You don't mention it in your comments.

soumith:
Regarding BLAS Handles, response from Nicolas(FB):
"They are fundamentally different resources with different lifetimes.
Atm, you can run multiple cublas calls with multiple streams on a single handle or only use a single stream across multiple handles.
I don’t have a strong argument against merging them but there may be performance / memory consumption implications of always forcing a single stream to a single handle.
The problem is I can’t think of a good test that would prove or disprove this.

Running mxm (513x513x513): 53 iterations (parallel over streams),  1 batches, GReductions(virtual fmas)/s = 634.47793  time = 11.28ms
Running mxm (513x513x513): 53 iterations (parallel over streams),  1 batches, GReductions(virtual fmas)/s = 785.07805  time = 9.11ms
Running mxm (513x513x513): 53 iterations (parallel over streams),  1 batches, GReductions(virtual fmas)/s = 931.75268  time = 7.68ms
Running mxm (513x513x513): 53 iterations (parallel over streams),  1 batches, GReductions(virtual fmas)/s = 934.15645  time = 7.66ms

These are simple perf results for respectively 1x1, 1x4, 4x1 and 4x4 handles x streams
If you have only 1 handle it’s better to have 4 streams.
But then it’s even better to have 4 handles and there is no significant gain from 4x4 over 4x1."

soumith:
I'm going to break these further in the future. It's just that we've gone too far from HEAD, and I wanted to just put the PR out there.

soumith:
The THCDeviceTensor* is an isolated subset of facebook::cuda( https://github.com/facebook/fbcuda ). We isolated the subset needed for the reduction kernels, and the Volumetric Max/Avgpooling kernels, and named them THCDevice* so that we dont have a dependency on fbcuda on the core of cutorch.
This is approximately what we discussed with @akfidjeland and @koraykv back in that email thread a couple of months ago.

FFI.lua
```diff
@@ -4,16 +4,28 @@ if ok then
    local cdefs = [[
 typedef struct CUstream_st *cudaStream_t;
 
+struct cublasContext;
+typedef struct cublasContext *cublasHandle_t;
+typedef struct CUhandle_st *cublasHandle_t;
+
+typedef struct _THCCudaResourcesPerDevice {
+  cudaStream_t* streams;
+  cublasHandle_t* blasHandles;
```
dominikgrewe: This doesn't match the definition in THCGeneral.h

lib/THC/THCGeneral.h
```diff
@@ -35,37 +35,45 @@
...
+
 
 /* Global state to be held in the cutorch table. */
 typedef struct THCState
 {
   struct THCRNGState* rngState;
-  struct THCBlasState* blasState;
   struct cudaDeviceProp* deviceProperties;
-  /* Convenience reference to the current stream in use */
+  /* Convenience reference to the current stream/handle/plan in use */
```
dominikgrewe: What do you mean by "plan"?

FFI.lua
soumith: cc: @nicolasvasilache

lib/THC/THCGeneral.h
soumith: cc: @nicolasvasilache

soumith:
cc: @nicolasvasilache please take a look at @dominikgrewe 's comments.

lib/THC/THCGeneral.h
soumith: I think plan probably means FFT plan.

lib/THC/THCGeneral.c
```diff
@@ -212,15 +341,33 @@ void THCState_setStream(THCState *state, int device, int stream)
...
-  state->currentStream = state->streamsPerDevice[device][stream];
+  state->currentStream =
+    THCState_getDeviceStream(state, device, stream);
   state->currentPerDeviceStream = stream;
-  THCudaBlas_setStream(state, device, state->currentStream);
+  THCublasCheck(cublasSetStream(state->currentBlasHandle,
+                                state->currentStream));
+}
+
+void THCState_setBlasHandle(THCState *state, int device, int handle)
```
dominikgrewe: You only really want to call this method if `device` is the current device, right? Otherwise `currentBlasHandle` holds a handle for a device other than the one we're currently using.

lib/THC/THCGeneral.h
nicolasvasilache: Yes, a plan is a cufft plan. The comment is outdated, a cufft plan has to be handled differently because it is problem size dependent unlike streams and handles. 

s:stream/handle/plan:stream/handle:

lib/THC/THCGeneral.h
```diff
@@ -35,37 +35,45 @@
...
-struct THCBlasState;
+
+typedef struct _THCCudaResourcesPerDevice {
+  cudaStream_t* streams;
+  cublasHandle_t* blasHandles;
+  /* Size of scratch space per each stream on this device available */
+  size_t scratchSpacePerStream;
+  /* Device-resident scratch space per stream, used for global memory
+     reduction kernels. */
+  void** devScratchSpacePerStream;
```
nicolasvasilache: Not sure where these scratchpads come from. cc Jeff: @wickedfoo

wickedfoo: I added that for the global reduction kernels. I follow the same strategy as Thrust: if the reduction is small, it's done in one block. If the reduction is large, two kernel launches/passes are used: each block does a reduction and writes back to a global memory scratch space, and a second kernel is launched to reduce from that.

Thrust allocates and deallocates the scratch buffer for every call, whereas I keep the scratch buffer space resident. The scratch space is only ~# of SMs * 4 * sizeof(float) bytes per stream, so relatively small.

This global memory scratch space can be used for other purposes too; anywhere where you need to perform two passes of work with not that much data needing to remain resident between the two passes.

It's managed here because we need independent space for each stream.

Also, I should convert logicalall and logicalany to this new reduction stuff as well, since those were added after I added the reduction stuff locally.

lib/THC/THCGeneral.c
nicolasvasilache: Correct, this is how it is used atm but it should be enforced (by having only a THCState_setBlasHandleForCurrentDevice function).

FFI.lua
wickedfoo: the scratch space needs to be added here

wickedfoo:
@soumith I'll send out a diff locally fixing the FFI and converting logicalany/logicalall that you can pull into this.

dominikgrewe:
As far as I can tell, THCDeviceTensor is not used in this PR (the reduce kernels don't use it). Can you factor it out into a separate PR? It should be fairly straightforward and we'd keep the commit history a bit saner.

soumith:
sure i'll do that,

hughperkins:
Hi, random comment, are you sure you need the `init` value?  I guess you can do things like?  :

```
  float r = modifyOp(in.data[inOffset]);
  inOffset += reductionStride;
  for (IndexType i = 1; i < reductionSize; ++i) {
    r = reduceOp(r, modifyOp(in.data[inOffset]));
    inOffset += reductionStride;
  }
```

wickedfoo:
@hughperkins To take the two parts in turn:

1. Needing init. I think I can remove this. The loops are block strided, so the block would always have to be # of threads >= totalElements in order for that to work. This I could do, but would have to check that in a couple of places, and there's this possible fragile dependency between the minimum number of elements and the block size. I'll look into that.

2. Simplifying the loop. There's this piece of code that occurs in various places:

```
    const IndexType inOffset = IndexToOffset<IndexType, ADims>::get(i, in);
```

the loop iterates in linear index order (i.e., [0, numElements -1]), which is converted to a real byte offset in a multi-dimensional array (with arbitrary strides/holes) using the appropriate math. You could be iterating over 5-d slices of a 6-d array, reducing over the 3rd dimension. It can't just be += reductionStride since the slice being reduced doesn't necessarily have a constant stride. This should compile down to effectively that in the case that the slice being reduced does have a constant stride, though.

lib/THC/THCTensorMasked.cu
```diff
@@ -44,11 +42,91 @@ void THCudaTensor_maskedFill(THCState* state,
...
+        maskPrefixSum(ps) {
+  }
+
+  __device__ __forceinline__ void operator()(float* out, float* mask) {
+    // Really mask should be `0` or `1` but we can't propagate errors here.
+    if (*mask != 0.0f) {
+      // We've already checked that this offset is <= 2^24, so this is ok.
+      int srcOffset = (int) (mask - baseMask);
+      *out = src[(int) maskPrefixSum[srcOffset]];
+    }
```
dominikgrewe: Can we not use `THCudaTensor_pointwiseApply3` here and pass in `maskPrefixSum` as an argument? Then we wouldn't need to calculate the offset.

lib/THC/THCGeneral.c
```diff
@@ -25,21 +28,33 @@ void THCudaInit(THCState* state)
...
+    res->blasHandles = NULL;
+
+    /* The scratch space that we want to have available per each device is
+       based on the number of SMs available per device */
+    int numSM = state->deviceProperties[i].multiProcessorCount;
+    size_t sizePerStream = numSM * GLOBAL_SCRATCH_SPACE_PER_SM_STREAM;
+    res->scratchSpacePerStream = sizePerStream;
+
+    /* Allocate scratch space for each stream */
+    res->devScratchSpacePerStream = malloc(sizeof(void*));
```
dominikgrewe: Please cast this to `void**` to avoid compiler warnings.

lib/THC/THCGeneral.c
```diff
@@ -48,32 +63,52 @@ void THCudaInit(THCState* state)
...
-
     /* Free Torch-defined streams (0 is the default stream) */
     for (int stream = 1; stream <= state->numUserStreams; ++stream) {
-      THCudaCheck(cudaStreamDestroy(state->streamsPerDevice[dev][stream]));
+      THCudaCheck(cudaStreamDestroy(
+                    THCState_getDeviceStream(state, dev, stream)));
+    }
+    /* Free Torch-defined handles (0 is NULL for consistency with streams API) */
+    for (int handle = 1; handle <= state->numUserBlasHandles; ++handle) {
+      THCudaCheck(cublasDestroy(
```
dominikgrewe: This should be THCublasCheck

lib/THC/THCGeneral.c
```diff
@@ -129,37 +164,85 @@ void THCState_reserveStreams(THCState* state, int numStreams)
...
     THCudaCheck(cudaSetDevice(dev));
 
     /* +1 for the default stream as well */
-    newStreams[dev] =
+    cudaStream_t* newStreams =
       (cudaStream_t*) malloc((numStreams + 1) * sizeof(cudaStream_t));
 
-    /* Copy over old streams
+    void** newScratchSpace =
+      malloc((numStreams + 1) * sizeof(void*));
```
dominikgrewe: cast to `void**`

lib/THC/THCGeneral.c
```diff
@@ -129,37 +164,85 @@ void THCState_reserveStreams(THCState* state, int numStreams)
...
+       (0 is NULL, 1 ... numUserBlasHandles are rest) */
+    newBlasHandles[0] = NULL;
+    for (int hndl = 1; hndl <= state->numUserBlasHandles; ++hndl) {
+      newBlasHandles[hndl] = THCState_getDeviceBlasHandle(state, dev, hndl);
+    }
+
+    /* Allocate new handles */
+    for (int hndl = state->numUserBlasHandles + 1; hndl <= numBlasHandles; ++hndl) {
+      newBlasHandles[hndl] = NULL;
+      THCudaCheck(cublasCreate(newBlasHandles + hndl));
```
dominikgrewe: use `THCublasCheck`

dominikgrewe:
@soumith I just added a few more comments. Once those ones and the previous ones have been addressed please let me have another quick look, but it should be good to go then. Thanks.

hughperkins:
Per removing init being complicated because ... I guess because some blocks are empty, so that would leave the reduced value of the block being undefined, is that what you are saying?  So, but still, I guess the init value is a function only of the choice reduce operator, and could be a constant built into the reduce operator class?  I suppose that in the specific of using the `thrust` operators, that this isnt possible, since we cannot modify the thrust operator classes?

lib/THC/THCTensorMasked.cu
wickedfoo: yes, looks like it :)

hughperkins:
I get the following compile warning, using yesterday's goodies2:

```
/home/user/git/cltorch/lib/THCl/THClReduceAll.h:38:19: warning: comparison between signed and unsigned integer expressions [-Wsign-compare]
   if (numBlocks > scratchSpace) {
```
Looking at the original code:
```
inline long getTwoPassBlocks(THCState* state, long elements) {
  long numBlocks = THCCeilDiv(elements, THC_REDUCE_ALL_BLOCK_SIZE);

  // We can only have as many blocks as there is scratch space
  size_t scratchSpace =
    THCState_getCurrentDeviceScratchSpaceSize(state) / sizeof(float);
  THAssert(scratchSpace > 0);

  if (numBlocks > scratchSpace) {
    numBlocks = scratchSpace;
  }

  return numBlocks;
}
```
... looks like numBlocks is long, ie signed, but scratchSpace is size_t, ie unsigned?

hughperkins:
>As far as I can tell, THCDeviceTensor is not used in this PR (the reduce kernels don't use it). Can you factor it out into a separate PR? It should be fairly straightforward and we'd keep the commit history a bit saner.

eg CeilDiv is needed by THC[l]ReduceAll.cl/cuh currently, and CeilDiv is currently part of THC[l]DeviceTensor

hughperkins:
Personal observation: it would be easier to find functions if they are prefixed with the module name :-)  eg, in [THCReduceAll](https://github.com/torch/cutorch/blob/goodies2/lib/THC/THCReduceAll.cuh), there is a method `THCudaTensor_reduceAll` , which I would logically hunt for in THCTensor.  (At least, I spend a lot of time doing `grep` to find functions, since their name gives no indication of where to find them?)

dominikgrewe:
>> As far as I can tell, THCDeviceTensor is not used in this PR (the reduce kernels don't use it). Can you factor it out into a separate PR? It should be fairly straightforward and we'd keep the commit history a bit saner.

> eg CeilDiv is needed by THC[l]ReduceAll.cl/cuh currently, and CeilDiv is currently part of THC[l]DeviceTensor

Ah, I didn't see that it was part of the THCDeviceTensor code. Good catch, thanks!
It's unrelated to the rest of the THCDeviceTensor code though, so we should probably put it into a different file anyway.

dominikgrewe:
> Personal observation: it would be easier to find functions if they are prefixed with the module name :-) eg, in THCReduceAll, there is a method THCudaTensor_reduceAll , which I would logically hunt for in THCTensor. (At least, I spend a lot of time doing grep to find functions, since their name gives no indication of where to find them?)

I think all functions operating on tensors (and that's the vast majority) start with THCudaTensor, apart from "special" ones like methods wrapping cuBLAS. So maybe files like THCReduceAll should be renamed to THCTensorReduceAll? That would be more consistent with THCTensorMath etc and should hopefully make it fairly obvious where to look for functions like `THCudaTensor_reduceAll`.

Prefixing method names according to the files they're defined in means that users of THC need to know which function is defined where, which doesn't seem very sensible.

hughperkins:
Question: why do we pass get_local_size(0) into reduceBlock in THClTensor_reduceContigDim, rather than 'reductionSize'?  It seems that this works ok since all threads have r set to `init`, but if we changed it to reductoinSize, then we might not need the `init` var?  (I'm still a newbie to the code, so I'm posing a question really, rather than making a suggestion.)

wickedfoo:
> I think all functions operating on tensors (and that's the vast majority) start with THCudaTensor, apart from "special" ones like methods wrapping cuBLAS. So maybe files like THCReduceAll should be renamed to THCTensorReduceAll? That would be more consistent with THCTensorMath etc and should hopefully make it fairly obvious where to look for functions like THCudaTensor_reduceAll.

In my opinion, the distinction between reduceAll etc. is that they are not part of the C Torch tensor API (e.g., the same functions implemented in torch7 for CPU tensors). They are implementation utilities for implementing the C Torch CudaTensor API, and aren't meant for end-users of the C tensor api.

I can change if you want, or leave it up to @soumith.

wickedfoo:
They are C++ templates, not C API calls as well.

soumith:
@dominikgrewe please see https://github.com/torch/cutorch/pull/181 , @colesbury is taking this over to completion...

PR #169, opened 2015-06-03
Set exit code to 1 if tests failed.

d11:
This is convenient for e.g. continuous integration.

### DIFF
--- a/test/test.lua
+++ b/test/test.lua
@@ -2063,5 +2063,6 @@ end
 
 if runtests then
    cutorch.test()
+   os.exit(#tester.errors == 0 and 0 or 1)
 end
 return test


ajtulloch:
What if there are 256 failing tests?  :)

I'm kind of opposed to this - returning fine-grained information like this isn't really the point of process exit statuses.  

What CI framework uses the process return code? How does it handle reporting 255, 256, 257, and 258 failed tests?

http://www.gnu.org/software/libc/manual/html_node/Exit-Status.html

> When a program exits, it can return to the parent process a small amount of information about the cause of termination, using the exit status. This is a value between 0 and 255 that the exiting process passes as an argument to exit.

> Normally you should use the exit status to report very broad information about success or failure. You can’t provide a lot of detail about the reasons for the failure, and most parent processes would not want much detail anyway.

> There are conventions for what sorts of status values certain programs should return. The most common convention is simply 0 for success and 1 for failure. Programs that perform comparison use a different convention: they use status 1 to indicate a mismatch, and status 2 to indicate an inability to compare. Your program should follow an existing convention if an existing convention makes sense for it.

ajtulloch:
Like, why not exit with 1 if you have any failures, and 0 otherwise?

d11:
Sure, works for me! I only meant that it needs not to be zero when there are failures.

soumith:
if you squash your commits, this is good to go as well.

d11:
Thanks - squashed!

PR #174, opened 2015-06-09
THC standalone compilation

szagoruyko:
moved the CMake CUDA logic to lib/THC/CMakeLists.txt and added a check for standalone compilation as in TH.


soumith:
nice! thanks.

PR #176, opened 2015-06-10
Allow CudaTensors as indices

colesbury:
Allow indexFill, indexCopy, and index to accept CudaTensors as indices
instead of LongTensors. We already just do a copy from LongTensor to
CudaTensor in the cutorch implementations, so this changes doesn't
result in any additional loss in precision.

### DIFF
--- a/torch/generic/Tensor.c
+++ b/torch/generic/Tensor.c
@@ -383,31 +383,38 @@ static int torch_Tensor_(indexSelect)(lua_State *L)
 {
   THCState *state = cutorch_getstate(L);
   int narg = lua_gettop(L);
-  THTensor *tensor, *src;
-  THLongTensor *index;
+  THTensor *tensor, *src, *index;
+  THLongTensor *longIndex;
   int dim;
   if (narg == 3)
   {
     tensor = THTensor_(new)(state);
     src = luaT_checkudata(L, 1, torch_Tensor);
     dim = luaL_checkint(L, 2) - 1;
-    index = luaT_checkudata(L, 3, "torch.LongTensor");
+    index = luaT_toudata(L, 3, torch_Tensor);
+    longIndex = luaT_toudata(L, 3, "torch.LongTensor");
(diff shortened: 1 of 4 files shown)


torch/generic/Tensor.c
```diff
@@ -383,31 +383,38 @@ static int torch_Tensor_(indexSelect)(lua_State *L)
...
-    index = luaT_checkudata(L, 4, "torch.LongTensor");
+    index = luaT_toudata(L, 4, torch_Tensor);
+    longIndex = luaT_toudata(L, 4, "torch.LongTensor");
+    if (!index && !longIndex) luaT_typerror(L, 4, "Tensor | LongTensor");
     tensor = luaT_checkudata(L,1,torch_Tensor);
   }
   else
   {
-    luaL_error(L,"Tensor, number, LongTensor | Tensor, Tensor, number, LongTensor expected");
+    luaL_error(L,"Tensor, number, Tensor | LongTensor expected");
```
dominikgrewe: This should also mention the 4-argument option. How about something like this:

```lua
luaL_error(L, "[Tensor,] Tensor, number, Tensor | LongTensor expected");
```

dominikgrewe:
Thanks! Would you mind extending the indexSelect etc tests with an example of using a CudaTensor for the indices?

colesbury:
I've updated the error message and added a test for each of indexSelect, indexCopy, and indexFill.

I originally tried to extend the tests to run with both CudaTensor and LongTensor for indices, but it was tough to get that to work with compareFloatAndCuda[Tensor]. Anyways, I think this covers all the code paths, since the LongTensor and CudaTensor mostly overlap.

dominikgrewe:
Yes, I think that's fine. Thanks a lot.

hughperkins:
Nice!

hughperkins:
Ported (mostly) to cltorch https://github.com/hughperkins/cltorch/commit/3fa2a0976827714ec3971bb4210485ede137f3ab

PR #178, opened 2015-06-11
Optimized indexSelect kernel for contiguous inputs

colesbury:
Adds an optimized indexSelect kernel for contiguous inputs indexed in
the first dimension. This will remove the need for a separate CUDA
LookupTable forward pass, since it can just use indexSelect.

### DIFF
--- a/lib/THC/THCTensorIndex.cu
+++ b/lib/THC/THCTensorIndex.cu
@@ -117,6 +117,7 @@ void THCudaTensor_indexCopy(THCState *state, THCudaTensor *res_, int dim, THCuda
   THArgCheck(nIndex == src->size[dim], 4, "length of src.size[dim] is not equal to length of indices");
 
   src = THCudaTensor_newContiguous(state, src);
+  indices = THCudaTensor_newContiguous(state, indices);
 
   nRes = THCudaTensor_nElement(state, res_);
   dim3 nthreads(16, 16);
@@ -133,6 +134,7 @@ void THCudaTensor_indexCopy(THCState *state, THCudaTensor *res_, int dim, THCuda
   );
 
   THCudaCheck(cudaFree(stride_));
+  THCudaTensor_free(state, indices);
   THCudaTensor_free(state, src);
 }
 
@@ -160,6 +162,7 @@ void THCudaTensor_indexFill(THCState *state, THCudaTensor *res_, int dim, THCuda
   THArgCheck(res_->nDimension > 0, 2, "Source tensor is empty");
 
   nRes = THCudaTensor_nElement(state, res_) / res_->size[dim] * nIndex;
+  indices = THCudaTensor_newContiguous(state, indices);
 
   dim3 nthreads(16, 16);
   dim3 nblocks(ceil((float)nRes / nIndex / (16*16)));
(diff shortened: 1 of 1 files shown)


dominikgrewe:
Have you tried to use cudaMemcpy? It probably depends on the inputs, but I could image it being faster for copying a small number of large slices. Haven't tried it myself though.

lib/THC/THCTensorIndex.cu
```diff
@@ -245,6 +266,24 @@ void THCudaTensor_indexSelect(THCState *state, THCudaTensor *res_, THCudaTensor
   THCudaTensor_resize(state, res_, newSize, NULL);
   THLongStorage_free(newSize);
 
+  if (THCudaTensor_isContiguous(state, src) && dim == 0)
```
dominikgrewe: The other tensors have to be contiguous too, don't they?

lib/THC/THCTensorIndex.cu
```diff
@@ -175,6 +175,26 @@ void THCudaTensor_indexFill(THCState *state, THCudaTensor *res_, int dim, THCuda
   THCudaCheck(cudaFree(stride_));
 }
 
+__global__ void THCudaTensor_kernel_indexSelect_contiguous(
```
dominikgrewe: Could you add a comment explaining how you're mapping the work to the thread blocks? Will make it easier to maintain the code in future.

colesbury:
I've added a comment about how the work is divided among blocks and threads.

I've made the sure the indices tensor is contiguous. I should have done this in the previous patch, as it affects all the existing indexFill,indexSelect,indexCopy kernels as well.

The performance is comparable to Tensor:copy for copying a small number of large slices. I got 90.4 GB/s for 8 slices of 10 million elements vs 88.5 GB/s for Tensor:copy. For small inputs, the time is about 7.7µs per indexSelect.

While benchmarking, I realized that large inputs can exceed the max CUDA size of a dimension of a dim3. This could happen if a slice is larger than 2^23 (8.3 million) or the number of indices is larger than 2^18 (262,144). This is fixed now.

lib/THC/THCTensorIndex.cu
```diff
@@ -245,6 +274,29 @@ void THCudaTensor_indexSelect(THCState *state, THCudaTensor *res_, THCudaTensor
   THCudaTensor_resize(state, res_, newSize, NULL);
   THLongStorage_free(newSize);
 
+  indices = THCudaTensor_newContiguous(state, indices);
+
+  if (THCudaTensor_isContiguous(state, src) && dim == 0)
```
dominikgrewe: You should also check that `res_` is contiguous. `THCudaTensor_resize` doesn't guarantee that.

lib/THC/THCTensorIndex.cu
```diff
@@ -173,6 +176,31 @@ void THCudaTensor_indexFill(THCState *state, THCudaTensor *res_, int dim, THCuda
...
+
+__global__ void THCudaTensor_kernel_indexSelect_contiguous(
+  float *tensor, float *src, long stride, float *index, long idxSize)
+{
+  // In the typical case, each block of 128 threads handles a 4x128
+  // section of the output with each warp handling a single 1x128 row.
+  // The outer loops handle inputs larger than 4*65535 or strides larger
+  // than 128*65535.
+
+  for (int idx = blockIdx.x * 4 + threadIdx.y; idx < idxSize; idx += 4*65535) {
```
dominikgrewe: Can you use `blockDim.x` etc instead of constants? It makes the code more easily readable and less error-prone.

Issue #165, opened 2015-05-27
THCRandom does not work with streams


wickedfoo:
THCRNGState's gen->gen_states are thread-safe as per curand within a block, but are (likely?) not safe when being accessed concurrently from different streams. 

We need to allocate a chunk of MAX_NUM_BLOCKS gen_states per each stream/each GPU currently in use, not one per each GPU, and use the appropriate one based on the currently active stream.

dominikgrewe:
@soumith Was wondering if you guys happen to have a fix for this already. Otherwise I'll prepare one.

wickedfoo:
Not on our end, go ahead.

FYI I fixed the Thrust + streams issue locally, though that requires moving cutorch/cunn/etc. to CUDA 7.

soumith:
the CUDA 7 is not a hard-limit. We have #ifdef guards. But without CUDA 7, thrust stuff wont respect streams.

wickedfoo:
I think you should officially require CUDA 7 at some point though :)

Otherwise, you shouldn't use streams at all, since it's highly likely you will get garbage or improper results.

wickedfoo:
@dominikgrewe I can get to this next if you haven't yet.

dominikgrewe:
I've been meaning to do it for a while but haven't gotten round to it yet. If you have time to do it, that'd be great!

PR #188, opened 2015-06-24
THCRandom: fixed possibly wrong pointers coming from a saved state

mys007:
Currently, a saved cutorch rng-state is not transferable between different machines or even different GPUs. The state can be loaded but follow-up operations end up with "cuda runtime error : an illegal memory access was encountered". This is because the saved state contains GPU-pointers to mtgp32_kernel_params_t, which are then of course not valid anymore. The suggest patch rewrites them with the valid memory addresses. 

The bug can be demonstrated by two scripts:
```
require 'cunn'
cutorch.manualSeed(1)
A=torch.CudaTensor(10):normal()
print(A)
torch.save('cc.bin',cutorch.getRNGState())
A=torch.CudaTensor(10):normal()
print(A)
```
```
require 'cunn'
cutorch.manualSeed(10)
A=torch.CudaTensor(10):normal()
print(A)
st=torch.load('cc.bin')
cutorch.setRNGState(st)
A=torch.CudaTensor(10):normal()
print(A)
```

### DIFF
--- a/lib/THC/THCTensorRandom.cu
+++ b/lib/THC/THCTensorRandom.cu
@@ -147,6 +147,11 @@ __host__ unsigned long THCRandom_initialSeed(THCState* state)
 
 __host__ void THCRandom_getRNGState(THCState* state, THByteTensor *rng_state)
 {
+  if (state->rngState->current_gen == NULL)
+  {
+    THError("Random number generators have not been initialized.");
+  }
+  
   // The RNG state comprises the MTPG32 states and the seed.
   static const size_t states_size = MAX_NUM_BLOCKS * sizeof(curandStateMtgp32);
   static const size_t seed_size = sizeof(unsigned long);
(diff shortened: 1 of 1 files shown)


dominikgrewe:
The two scripts seem to work fine for me; at least if I understand them correctly. The second tensor is the same each time. What are you getting?

First script:
```
-0.5472
-0.7301
 0.5129
-0.7344
 0.8185
 0.3763
 0.2013
 0.6852
 0.1692
 1.0536
[torch.CudaTensor of size 10]

 1.2795
 0.7561
 1.5369
 1.2223
-1.1212
-0.8355
-0.3413
-0.0317
-0.7776
-1.6136
[torch.CudaTensor of size 10]
```

Second script:
```
-0.6813
 1.0871
-1.3523
 1.8390
 0.4403
-0.5530
 0.1154
-0.4803
 0.0994
 0.3822
[torch.CudaTensor of size 10]

 1.2795
 0.7561
 1.5369
 1.2223
-1.1212
-0.8355
-0.3413
-0.0317
-0.7776
-1.6136
[torch.CudaTensor of size 10]
```

mys007:
That is the correct result. But have tried to produce the file cc.bin on a different machine?

dominikgrewe:
I see, good point. Didn't realise it copies pointers.

I think it'd be slightly cleaner if you set the kernel_params pointers after copying the generator state. It avoids having to modify the input.

mys007:
True. But for that, do I have to write a new kernel... I'd rather clone the function argument...

dominikgrewe:
Should be a fairly short and simple kernel though :)

Alternatively, we should at least do either of the following:

1) Cloning the state before modifying it (somewhat inefficient for the typical use case).
2) Document clearly that the state is being modified.

mys007:
Hi, so I wrote it as a kernel.

lib/THC/THCTensorRandom.cu
```diff
@@ -172,15 +177,28 @@ __host__ void THCRandom_getRNGState(THCState* state, THByteTensor *rng_state)
...
   static const size_t states_size = MAX_NUM_BLOCKS * sizeof(curandStateMtgp32);
   static const size_t seed_size = sizeof(unsigned long);
   static const size_t total_size = states_size + seed_size;
   THArgCheck(THByteTensor_nElement(rng_state) == total_size, 1, "RNG state is wrong size");
   THArgCheck(THByteTensor_isContiguous(rng_state), 1, "RNG state must be contiguous");
+  
   THCudaCheck(cudaMemcpy(state->rngState->current_gen->gen_states, THByteTensor_data(rng_state),
                          states_size, cudaMemcpyHostToDevice));
+  set_rngstate_kernel<<<1, MAX_NUM_BLOCKS, 0, THCState_getCurrentStream(state)>>>(
+      state->rngState->current_gen->gen_states, state->rngState->current_gen->kernel_params);
```
dominikgrewe: I think we should synchronize after this call to avoid races between this kernel call and other RNG calls on different streams. We only have one RNG state per device, not per stream. Which, thinking about it, doesn't seem very safe... I couldn't find any information whether of not it's safe in the cuRAND docs though. Does anyone know more about this?

dominikgrewe:
Thanks! I just have one comment. Otherwise it looks good to me. If you could address the comment and squash your commits, it should be ready to merge.

mys007:
Shouldn't the sync be done before the kernel as well, for the same reason? By the way, the stream issue is already open at https://github.com/torch/cutorch/issues/165. Personally, I don't even know how to use more streams with cutorch:).

dominikgrewe:
You're right. I'd forgotten about that issue. Let's leave this PR as-is and then fix the actual issue separately.
Can you please squash your two commits?

mys007:
Done

dominikgrewe:
Thanks! There seem to be conflicts, can you rebase please?

mys007:
Done, sorry for that.

dominikgrewe:
No worries. Thanks a lot for the fix!

PR #192, opened 2015-07-03
Always use events for synchronization when copying across devices.

dominikgrewe:
When using the default stream, cudaMemcpyAsync guarantees that the copy is synchronized with respect to both devices for a peer-to-peer copy: http://docs.nvidia.com/cuda/cuda-c-programming-guide/index.html#peer-to-peer-memory-copy
However, it doesn't state anything about copies when using non-default streams. To have consistent behaviour we should always use events to enforce synchronization when copying data across GPUs.


soumith:
good catch. yes, last time round when we discussed this code, there were no streams support in cutorch :)

PR #196, opened 2015-07-08
Fix cmake build errors on older systems

szagoruyko:
Should fix build errors like this:
```
-- Found CUDA: /usr/local/cuda-7.0 (found suitable version "7.0", required is "5.5")
CMake Error at /home/corona/torch/install/share/cmake/torch/TorchExports.cmake:16 (ADD_LIBRARY):
  add_library cannot create imported target "TH" because another target with
  the same name already exists.
Call Stack (most recent call first):
  /home/corona/torch/install/share/cmake/torch/TorchConfig.cmake:29 (INCLUDE)
  lib/THC/CMakeLists.txt:3 (FIND_PACKAGE)
```


soumith:
i could never reproduce this. I guess it was an older system or older cmake issue. i'll bump up the distro commit as well.

PR #206, opened 2015-07-21
CUDA implementations for scatter & gather.

dominikgrewe:


dominikgrewe:
The implementation is very similar to @hughperkins's code in cltorch. Two things are worth pointing out:

1) I couldn't use the regular `TensorInfo` object because it flattens dimensions when possible, where as I need to keep the size and stride information as they are. I therefore added a `SimpleTensorInfo` struct.

2) I use template specialization for a small number of dimensions. On a simple benchmark (`index size = 1000 x 10 x 100` and `dim = 2`) this reduces the runtime by 33% for gather and 42% for scatter.

soumith:
thanks dominik!

hughperkins:
@dominikgrewe just out of personal curiosity, how do these structs arrive in the kernel in CUDA?  In OpenCL, I've been allocating a memory buffer for them, just like for the underlying data itself, and copy it across, which is quite slow.  On CUDA, do they just get sent with the rest of the kernel parameters all together as a single blob?

wickedfoo:
@hughperkins they arrive on the GPU effectively as a memcpy of sizeof(args...); the structs are POD and that just works. Though, I wonder what happens for big-endian platforms and little-endian platforms when pointers are built in. Maybe the compiler in the kernel call (as part of parsing `<<<>>>`) handles converting whatever needs converting when building the blob.

PR #208, opened 2015-07-23
Merge facebook changes

adamlerer:
Some facebook changes by @wickedfoo
2) `torch.multinomial` support added to `CudaTensor`
3) integrated thurst stream support that was added in CUDA 7

### DIFF
--- a/lib/THC/THCTensorIndex.cu
+++ b/lib/THC/THCTensorIndex.cu
@@ -5,17 +5,8 @@
 #include "THCTensorRandom.h"
 #include "THCApply.cuh"
 #include "THCReduce.cuh"
-
-#include <algorithm>
-#include <thrust/device_ptr.h>
-#include <thrust/fill.h>
-#include <thrust/functional.h>
-#include <thrust/reduce.h>
-#include <thrust/inner_product.h>
-
-#ifndef DIVUP
-#define DIVUP(x, y) (((x) + (y) - 1) / (y))
-#endif
+#include "THCDeviceUtils.cuh"
+#include <algorithm> // for std::min
 
 __global__ void THCudaTensor_kernel_indexFill(
    float *tensor, long* stride, float *index, long src_nDim,
@@ -286,8 +277,8 @@ void THCudaTensor_indexSelect(THCState *state, THCudaTensor *res_, THCudaTensor 
   {
     long stride = src->stride[0];
 
-    int blockX = std::min(DIVUP(nIndex, 4), 65535L);
-    int blockY = std::min(DIVUP(stride, 128), 65535L);
+    int blockX = std::min(THCCeilDiv(nIndex, 4L), 65535L);
+    int blockY = std::min(THCCeilDiv(stride, 128L), 65535L);
 
     dim3 nthreads(32, 4);
     dim3 nblocks(blockX, blockY);
--- a/lib/THC/THCTensorRandom.cu
+++ b/lib/THC/THCTensorRandom.cu
@@ -1,5 +1,9 @@
 #include "THCTensorRandom.h"
+#include "THCDeviceUtils.cuh"
 #include "THCGeneral.h"
+#include "THCTensorCopy.h"
+#include "THCTensorMath.h"
+#include "THCReduceApplyUtils.cuh"
 
 #include <thrust/functional.h>
 #include <curand.h>
@@ -9,10 +13,6 @@
 
 #define MAX_NUM_BLOCKS 64
 #define BLOCK_SIZE 256
-
-#ifndef DIVUP
-#define DIVUP(x, y) (((x) + (y) - 1) / (y))
-#endif
 
 /* Sets up generator. Allocates but does not create the generator states. */
 __host__ void initializeGenerator(Generator* gen)
@@ -151,7 +151,7 @@ __host__ void THCRandom_getRNGState(THCState* state, THByteTensor *rng_state)
   {
     THError("Random number generators have not been initialized.");
   }
-  
+
   // The RNG state comprises the MTPG32 states and the seed.
   static const size_t states_size = MAX_NUM_BLOCKS * sizeof(curandStateMtgp32);
   static const size_t seed_size = sizeof(unsigned long);
@@ -175,13 +175,13 @@ __host__ void THCRandom_setRNGState(THCState* state, THByteTensor *rng_state)
   {
     THError("Random number generators have not been initialized.");
   }
-  
+
   static const size_t states_size = MAX_NUM_BLOCKS * sizeof(curandStateMtgp32);
   static const size_t seed_size = sizeof(unsigned long);
   static const size_t total_size = states_size + seed_size;
   THArgCheck(THByteTensor_nElement(rng_state) == total_size, 1, "RNG state is wrong size");
   THArgCheck(THByteTensor_isContiguous(rng_state), 1, "RNG state must be contiguous");
-  
+
   THCudaCheck(cudaMemcpy(state->rngState->current_gen->gen_states, THByteTensor_data(rng_state),
                          states_size, cudaMemcpyHostToDevice));
   set_rngstate_kernel<<<1, MAX_NUM_BLOCKS, 0, THCState_getCurrentStream(state)>>>(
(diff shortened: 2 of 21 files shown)


szagoruyko:
tests fail sometimes:
```
seed: 	1437644687
Running 121 tests
________________________________________________________________________________________________________________________*  ==> Done Completed 40428 asserts in 121 tests with 4 errors
--------------------------------------------------------------------------------
multinomial_with_replacement
sampled an index with zero probability
 BOOL violation   condition=false
	...ruys/rocks/distro/install/share/lua/5.1/torch/Tester.lua:22: in function 'assert'
	...ruys/rocks/distro/install/share/lua/5.1/cutorch/test.lua:1650: in function 'v'
...
```

adamlerer:
Thanks @szagoruyko. Running with your seed, `cutorch.test(nil, 1437644687)` doesn't fail for me, but I can reproduce by running `cutorch.test('multinomial_with_replacement', 1437667171)`. Looking into it.

adamlerer:
Okay, multinomial-with-replacement bug fix and test improved.

adamlerer:
Removed the cutorch GC integration from this PR, it needs a bit more work and then I'll send out a separate PR for it.

lib/THC/THCTensorIndex.cu
```diff
@@ -10,7 +10,6 @@
 #include <thrust/fill.h>
 #include <thrust/functional.h>
 #include <thrust/reduce.h>
```
dominikgrewe: We don't seem to be using thrust at all in this file. Can you just remove all of the includes?

dominikgrewe:
There are a number of files that don't use thrust at all but include various thrust headers. Would you mind cleaning those up?

Also, THCStorage.cu has a call to thrust::fill which needs to get a stream passed in.

test/test.lua
```diff
@@ -1634,6 +1634,93 @@ function test.multi_gpu_random()
...
+function test.multinomial_with_replacement()
+   for tries = 1, 10 do
+      local n_row = torch.random(10)
+      local n_col = 1 + torch.random(1000)
+
+      local prob_dist = torch.CudaTensor(n_row, n_col):uniform()
+      prob_dist:select(2, n_col):fill(0) --index n_col shouldn't be sampled
+      local n_sample = torch.random(n_col - 1)
+      local sample_indices = torch.multinomial(prob_dist, n_sample, true)
+      tester:assert(prob_dist:dim() == 2, "wrong number of prob_dist dimensions")
```
dominikgrewe: Why are you checking for prob_dist? You've created that yourself.
Wouldn't it be more useful to check that sample_indices has the right number of dimensions and the right size?

test/test.lua
```diff
@@ -1634,6 +1634,93 @@ function test.multi_gpu_random()
...
+      prob_dist:select(2, n_col):fill(0) --index n_col shouldn't be sampled
+      local n_sample = torch.random(n_col - 1)
+      local sample_indices = torch.multinomial(prob_dist, n_sample, false)
+      tester:assert(prob_dist:dim() == 2, "wrong number of prob_dist dimensions")
+      tester:assert(sample_indices:size(2) == n_sample, "wrong number of samples")
+
+      for i = 1, n_row do
+         local row_samples = {}
+         for j = 1, n_sample do
+            local sample_idx = sample_indices[{i,j}]
```
dominikgrewe: This will be rather slow. Each index triggers a memcpy from the GPU which has high latency. It should be faster to copy the entire tensor to the CPU and then do the checks on that.

lib/THC/THCTensorRandom.cu
```diff
@@ -359,4 +359,405 @@ THC_API void THCudaTensor_cauchy(THCState* state, THCudaTensor *self_, double me
...
+    inputSize == 1 ? THCudaTensor_size(state, prob_dist, 0) :
+    THCudaTensor_size(state, prob_dist, 1);
+
+  // Since the index tensor is float, numCategories cannot exceed max=
+  // float integer precision
+  THArgCheck(numCategoriesLong <= FLOAT32_MAX_CONSECUTIVE_INT, 2,
+             "number of categories cannot exceed 2^24");
+  int numCategories = (int) numCategoriesLong;
+
+  THArgCheck(n_sample > 0, 3, "cannot sample n_sample < 0 samples");
```
dominikgrewe: nit: "cannot sample __<=__ 0 samples"

lib/THC/THCTensorRandom.cu
```diff
@@ -359,4 +359,405 @@ THC_API void THCudaTensor_cauchy(THCState* state, THCudaTensor *self_, double me
...
+
+  if (!with_replacement) {
+    THArgCheck(n_sample <= numCategories, 2,
+               "cannot sample n_sample > prob_dist:size(1) samples without "
+               "replacement");
+  }
+
+  // Restructure data for 2d
+  if (inputSize == 1) {
+    THCudaTensor_resize2d(state, prob_dist, 1, numCategories);
```
dominikgrewe: This resize operation only works as expected if prob_dist is contiguous, because resize ignores the tensor's stride. So you should make it contiguous before you do this (you make it contiguous later anyway).

The below one is fine, because it's not an input.

lib/THC/THCTensorRandom.cu
```diff
@@ -359,4 +359,405 @@ THC_API void THCudaTensor_cauchy(THCState* state, THCudaTensor *self_, double me
...
+  // the random number generator state is fully updated.
+
+  // The block determines the distribution for which we generate a point
+  for (long curDist = blockIdx.x;
+       curDist < distributions;
+       curDist += gridDim.x) {
+    for (int sampleBase = threadIdx.y;
+         sampleBase < totalSamples; sampleBase += blockDim.y) {
+      // The warp determines the sample for which we generate a point
+      int sample = sampleBase + blockIdx.y;
```
dominikgrewe: Isn't `blockIdx.y` always 0? Your grid is 1D.

lib/THC/THCTensorRandom.cu
```diff
@@ -359,4 +359,405 @@ THC_API void THCudaTensor_cauchy(THCState* state, THCudaTensor *self_, double me
...
+
+__global__ void
+sampleMultinomialWithReplacement(curandStateMtgp32* state,
+                                 int totalSamples,
+                                 float* dest,
+                                 long distributions,
+                                 int categories,
+                                 float* normDistPrefixSum) {
+  // Binary search is by its nature warp divergent, so all threads
+  // in a warp compute the save value. It's important that all threads
```
dominikgrewe: What do you mean by "all threads in a warp compute the save value"? Only one thread in each warp actually computes a value, right?

Have you actually tried using more than one thread per warp? The depth of your binary search is always the same, isn't it? So the only divergence is in the `midVal < val` condition, which you could re-write using the ternary operator. Would be interesting to see if that buys you anything.

lib/THC/THCTensorRandom.cu
wickedfoo: I think this was a bad edit for a curand_uniform fix, but looking at this code it still looks broken. Should be
```
for (int sampleBase = 0; sampleBase < totalSamples; sampleBase += blockDim.y) {  
  int sample = sampleBase + threadIdx.y);
```

lib/THC/THCTensorRandom.cu
wickedfoo: The comment rotted I think. The code used to have all threads participate, but only one thread saved the value. Should be save -> same too.

Good idea, I'll try this.

dominikgrewe:
@adamlerer Did you have some time to address my comments? I'll look at your other pull requests now, but it'd be good to get this in too.

Regarding the possible optimisation in the `sampleMultinomialWithReplacement` kernel, it shouldn't prevent us from merging this PR.

adamlerer:
@wickedfoo is going to address these comments since these contributions are his. Thanks!

adamlerer:
@wickedfoo addressed the comments and I've amended the commit.

dominikgrewe:
Thanks. There's still a call to `thrust::fill` in THCStorage.cu which should be updated. Otherwise this looks good to me.

wickedfoo:
Adam I fixed that  particular issue the other week in a different internal
diff, you can just grab that snippet.

adamlerer:
Fixed, thanks Dominik!

dominikgrewe:
Thanks guys! Looks like there are merge conflict. Can you rebase?

soumith:
thanks a lot for taking this through completion adam!

PR #209, opened 2015-07-24
Making things work under windows

samehkhamis:
Some changes to get things working under Windows:

CMake defines _WIN32, not WIN32.
Werror is not supported under MSVC.
Explicitly linking with LUALIB.
Exporting functions explicitly.

Cheers.


soumith:
same comment as on cunn, can we make it work with windows without removing the:  -j$(getconf _NPROCESSORS_ONLN)

samehkhamis:
Hi Soumith,

I think that fixes it.

Cheers.

soumith:
thanks a lot!

Issue #213, opened 2015-07-31
mm/addmm fails with stride=0


avnermay:
require 'cutorch'
torch.setdefaulttensortype('torch.CudaTensor')
A = torch.randn(100,3200)
B = torch.randn(1,10):expand(100,10)
C = torch.randn(10,3200)
torch.mm(B,C) -- THIS LINE THROWS ERROR
A:addmm(0,1,B,C) -- THIS LINE DOES TOO

This is the error I am getting:
 ** On entry to SGEMM  parameter number 10 had an illegal value
Segmentation fault (core dumped)

soumith:
thanks. we fixed this for torch, I suppose I'll have to apply similar patches for cutorch. Will do now.

avnermay:
Great thanks a lot!

avnermay:
One additional comment: 
This issue was giving me trouble when doing a backward pass on a Linear node (line 66 of nn.Linear.lua: "self.gradInput:addmm(0, 1, gradOutput, self.weight)") -- Given that I have a decently sized code base, it was actually quite time consuming to figure out exactly what line of code was throwing this error.  It would be great if all BLAS/CUBLAS errors were logged to stderr with a stack trace.  Currently, the only thing that appears is the error message I copied above, without any info about what function was causing the error.

fmassa:
I'm preparing a patch for it. Just finishing some tests

soumith:
nice!

avnermay:
Awesome thanks so much.  You guys are fast!  

Any comment on the debugging/stack-trace issue?

avnermay:
(and one unrelated question: when does nn.oldSeed get set?)

soumith:
@avnermay nn.oldSeed gets set manually. it's just for backward compatibility (and reproducibility) with really really old scripts.

soumith:
wrt the debuggin/stack-trace issue, if we have an assertion, like THError or LuaL_check* functions, the stack trace gets printed. however, this was an error coming directly from cublas, which kinda caught us off-guard

avnermay:
Great thanks a lot -- if there are other calls into blas/cublas (in other places in Torch) that may result in similarly hard to interpret error messages, it would be great to fix those as well.

PR #214, opened 2015-07-31
Fix for gemm with zero strides

fmassa:
Fixes https://github.com/torch/cutorch/issues/213

There is still one issue though (also present in `torch`), when the output Tensor has zero stride and the good dimensions the results are not right.
This is probably due to an issue with `freeCopyTo`. Indeed, it calls `copy`, which doesn't work as expected when the destination Tensor has zero stride and the good dimensions, as discussed in https://github.com/torch/torch7/issues/289.


soumith:
thanks Francisco

PR #218, opened 2015-08-05
Add THCDeviceTensor

adamlerer:


hughperkins:
Name sounds interesting.  What does this do?

soumith:
it's a subset of [fbcuda](http://fbcuda.readthedocs.org/en/latest/reference.html). It allows one to do (among many other things) tensor indexing inside CUDA kernels without having to calculate ugly offsets based on size/stride.

lib/THC/THCDeviceTensor.cuh
```diff
@@ -0,0 +1,502 @@
...
+  __host__ __device__ __forceinline__
+  const detail::THCDeviceSubTensor<TensorType, Dim - 1, PtrTraits>
+    operator[](IndexT) const;
+
+  /// Returns the size of a given dimension, `[0, Dim - 1]`. No bounds checking.
+  __host__ __device__ __forceinline__ int getSize(int i) const {
+    return size_[i];
+  }
+
+  /// Returns the size of a given dimension, `[0, Dim - 1]`. No bounds checking.
```
dominikgrewe: size -> stride

lib/THC/THCDeviceTensor.cuh
```diff
@@ -0,0 +1,502 @@
...
+          int Dim,
+          typename IndexT,
+          template <typename U> class PtrTraits>
+class THCDeviceTensor;
+
+/// Type of a subspace of a tensor
+namespace detail {
+template <typename TensorType,
+          int SubDim,
+          template <typename U> class PtrTraits>
```
dominikgrewe: Is `PtrTraits` ever different from `TensorType::PtrTraits`?

lib/THC/THCDeviceTensor-inl.cuh
```diff
@@ -0,0 +1,424 @@
...
+#endif
+
+  union {
+    T vT;
+    int vInt;
+  } u;
+
+  u.vT = val;
+
+  cudaMemsetAsync(data(), u.vInt, numElements() * sizeof(T), stream);
```
dominikgrewe: If I understand `cudaMemset` correctly, this will only look at the lower 8 bits of the integer value and copy those `numElements() * sizeof(T)` times. So for data types with more than 8 bits this won't do what you intend to do.

lib/THC/THCDeviceTensor.cuh
wickedfoo: This was written against C++03, there is no TensorType::PtrTraits because you can't typedef a template. In C++11 you could say something like `template <typename U> using PtrTraits = PtrTraits<U>;` in THCDeviceTensor and use it that way.

Another C++03 option would have been to add an extra level of indirection, like

```
class THCDeviceTensor {
  ...
  template <typename U> struct UsingTraits { typedef PtrTraits<U> type; };
```

I just chose to template the sub-class on the traits in order to use it.

lib/THC/THCDeviceTensor-inl.cuh
wickedfoo: yes this is broken, don't know how this happened. will change to zero() to just clear it instead.

adamlerer:
Sorry about the delay Dominik. I think this is ready to be merged now.

dominikgrewe:
Thanks!

Do you have plans to merge THCDeviceTensor and THCudaTensor? It feels almost like we have two implementations of the same thing... Maybe with a move to templated tensors in TH and more types for cutorch tensors, we can consolidate the two implementations?

wickedfoo:
They serve different purposes in my eyes: THCudaTensor is meant to implement the Torch C tensor API.
THCDeviceTensor is meant for use in writing CUDA kernels, and can only be used for cases where the number of dimensions is statically known (the type includes the dimension). THCudaTensor does not contain any static information about the size of tensors.

wickedfoo:
A templated tensor implementation that doesn't have static knowledge of the # of dimensions would make sense to support different datatypes, but Torch long ago decided against that and instead is re-implementing C++ templates via C macros and/or massive code duplication in my opinion :(

dominikgrewe:
Good point re static dimensions.

There's hope that Torch will eventually switch to C++ templates: https://github.com/torch/torch7/blob/master/ROADMAP.md

PR #219, opened 2015-08-05
cutorch automatic GC

adamlerer:
This is the cutorch version of https://github.com/torch/torch7/pull/280

Since there's no _msize for cutorch, we're going to only keep track of heap usage by THCStorages, which take up most of the space.

### DIFF
--- a/init.lua
+++ b/init.lua
@@ -29,4 +29,6 @@ function cutorch.createCudaHostTensor(...)
   return torch.FloatTensor(storage, 1, size:storage())
 end
 
+cutorch.setHeapTracking(true)
+
 return cutorch
--- a/lib/THC/THCGeneral.c
+++ b/lib/THC/THCGeneral.c
@@ -53,7 +53,7 @@ void THCudaInit(THCState* state)
 
     /* Allocate scratch space for each stream */
     res->devScratchSpacePerStream = (void**) malloc(sizeof(void*));
-    THCudaCheck(cudaMalloc(&res->devScratchSpacePerStream[0],
+    THCudaCheck(THCudaMalloc(state, &res->devScratchSpacePerStream[0],
                            sizePerStream));
   }
 
@@ -71,6 +71,10 @@ void THCudaInit(THCState* state)
   THCState_reserveBlasHandles(state, 1);
   state->currentPerDeviceBlasHandle = 1;
   state->currentBlasHandle = THCState_getDeviceBlasHandle(state, device, 1);
+
+  state->cutorchGCFunction = NULL;
+  state->cutorchGCData = NULL;
+  state->heapSoftmax = 300000000; // 300MB, adjusted upward dynamically
 }
 
 void THCudaShutdown(THCState* state)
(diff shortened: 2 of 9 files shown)


lib/THC/THCGeneral.c
```diff
@@ -481,4 +481,58 @@ void __THCublasCheck(cublasStatus_t status, const char *file, const int line)
   }
 }
 
+static __thread void (*cutorchGCFunction)(void *data) = NULL;
+static __thread void *cutorchGCData = NULL;
+static __thread long heapSize = 0;
+static __thread long heapSoftmax = 300000000; // 300MB, adjusted upward dynamically
```
dominikgrewe: These should be part of the THCState rather than thread-local variables.

adamlerer:
There's actually a bigger problem which is that `heapSize` must not be thread-local at all, see https://github.com/torch/torch7/pull/324. The rest of it can be in `THCState`.

I'm having a bit of trouble testing threads with cutorch, when I run the example from https://github.com/torch/torch7/pull/324 with CudaTensors on master, I get weird errors, either

```
FATAL THREAD PANIC: (read) ...al/git/torch-distro/install/share/lua/5.1/torch/File.lua:266: read error: read 10 blocks instead of 542886064 at /data/users/alerer/git/torch7/lib/TH/THMemoryFile.c:323
```
or

```
FATAL THREAD PANIC: (read) ...al/git/torch-distro/install/share/lua/5.1/torch/File.lua:266: $ Torch: invalid memory size -- maybe an overflow? at /data/users/alerer/git/torch7/lib/TH/THGeneral.c:197
```

Tracking that down, then I'll update this.

dominikgrewe:
I don't understand why the heap size is now global but the the heap softmax is part of the THCState. Shouldn't they either both be global or both part of the THCState? Considering that the garbage collection function is part of the THCState, it seems consistent to have everything in the THCState.

dominikgrewe:
I saw your comment in the Torch PR. Didn't think about the case where a tensor is moved between states/threads. So it makes sense to keep the heap size global.

init.lua
```diff
@@ -29,4 +29,6 @@ function cutorch.createCudaHostTensor(...)
   return torch.FloatTensor(storage, 1, size:storage())
 end
 
+cutorch.setHeapTracking(true)
```
dominikgrewe: Do we want to enable it by default?

adamlerer:
In the torch7 gc discussion you said

"I agree that this should be disabled by default for now. In cutorch we have a good solution to push these counters into the Lua state with the THCState. I don't know how feasible it is to do something similar for torch."

so I assumed you didn't have the same reservations about cutorch. But if you prefer, I can disable it by default for a while until we get some more experience with it.

dominikgrewe:
Good point. I'd forgotten about that :)

Looks like there are merge conflicts. Can you rebase please?

adamlerer:
rebased and squashed commits

dominikgrewe:
Thanks!

PR #222, opened 2015-08-19
stream event fixes

soumith:
This patch was sent by Natalia Gimelshein from NVIDIA. Thanks Natalia.

Verbatim copy/paste of email:

I've noticed that there is incorrect implementation of cutorch stream synchronization functions, such as cutorch.streamWaitFor, cutorch.streamBarrier and cutorch.streamWaitForMultiDevice. One event per GPU is created, that it is recorded on each stream in the "synchronize with" list, and the cudaStreamWaitEvent is called in the waiting stream for this event. However, documentation
http://developer.download.nvidia.com/compute/cuda/4_1/rel/toolkit/docs/online/group__CUDART__EVENT_ga324d5ce3fbf46899b15e5e42ff9cfa5.html
says 
"If cudaEventRecord() has previously been called on event, then this call will overwrite any existing state in event. Any subsequent calls which examine the status of event will only examine the completion of this most recent call to cudaEventRecord()."
So the synchronization was performed only with the last stream on which the event was recorded. Tests passed because the work was enqueued to the streams in order, and last stream was done last. However, even when enqueueing work in order, there is no guarantee that the last stream will be done last, and besides, there is no requirement that work should be enqueued in order.
In addition, in cudaStreamWaitForMultiDevice the check should be if there is at least 1 stream to wait on, not 2, as currently.
I'm attaching a patch fixing these issues. I've also changed test.lua to enqueue work to streams in reverse order, so tests fail with current implementation, but pass with my patch.

### DIFF
--- a/init.c
+++ b/init.c
@@ -81,26 +81,28 @@ void checkAndCountListOfGPUStreamPairs(lua_State *L, THCState *state, int arg,
   lua_pop(L, 1);
 }
 
-void createSingleDeviceEvent(lua_State *L, THCState *state, int arg,
+int createSingleDeviceEvents(lua_State *L, THCState *state, int arg,
                              int device, cudaEvent_t* event)
 {
-  THCudaCheck(cudaEventCreateWithFlags(event, cudaEventDisableTiming));
 
   /* Push table to top */
   lua_pushvalue(L, arg);
 
   /* Record events */
   lua_pushnil(L);
+  int i = 0;
   while (lua_next(L, -2)) {
     int streamId = (int) lua_tonumber(L, -1);
     cudaStream_t streamWaitingOn =
       THCState_getDeviceStream(state, device, streamId);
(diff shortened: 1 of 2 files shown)


soumith:
cc: @dominikgrewe @wickedfoo

init.c
```diff
@@ -422,7 +430,7 @@ static int cutorch_streamWaitForMultiDevice(lua_State *L)
      for that GPU.
      -For (gpuWaiter, streamWaiter), wait on all of the above events.
   */
```
dominikgrewe: Comment needs updating.

dominikgrewe:
Looks good to me, but some comments now need to be updated. @wickedfoo should have the last word though, as he's more familiar with this code.

soumith:
Natalia addressed jeff's and dominik's comments with a revised patch that I pushed. Seems to look good. Merging.

Issue #216, opened 2015-08-01
pinned tensors need asynchronous copy routines


soumith:
This came up while discussing something with NVIDIA. 

Thanks to @dominikgrewe , we have the ability to allocate pinned tensors. However,

```lua
A = cutorch.createCudaHostTensor(3);
B = torch.CudaTensor(3);
B:copy(A); — this call should be asynchronous because A is pinned memory.
```

the example above is not asynchronous because our host->device copy routines are implemented via cudaMemcpy, and not cudaMemcpyAsync:

https://github.com/torch/cutorch/blob/master/torch/generic/Tensor.c#L573
https://github.com/torch/cutorch/blob/master/lib/THC/THCTensorCopy.c#L15

We need to create a mechanism to enable async copy on the pinned versions.

Should we do this directly in :copy() itself, or should we introduce a more painful :copyAsync()? Thoughts? cc: @wickedfoo @dominikgrewe

wickedfoo:
I'd for better or worse add copyAsync.

While :copy() is either synchronous or asynchronous with respect to the CPU
depending upon the location of arguments, it always is synchronous if
copying to or from the CPU. The fact that GPU to GPU copy is async is an
implementation detail that is not visible to the user unless they go off
the default stream, since the GPUs are effectively a black box that you
can't inspect unless you sync the GPU to CPU.

Similarly, the new :copyAsync() would be asynchronous when copying to or
from the CPU, and would map to old copy behavior for GPU to GPU copies. One
should be able to still copy synchronously to/from pinned memory, and
plugging pinned async copy into :copy() will remove the synchronous
ability. However, it should error if you're attempting an async copy
to/from CPU on non-pinned memory.

This is still weird admittedly. The only "clean" thing I think would be to
make :copy() synchronous with respect to the CPU everywhere, and
:copyAsync() async with respect to the CPU everywhere, but that would
require hunting down and changing more than half of all copy calls to
copyAsync, since otherwise GPU to GPU copy() would cudaStreamSynchronize or
event wait on the CPU to the GPU copy completion.

However the bigger problem is how one deals with the completion of the
async pinned memory copy on the CPU side? If you want to use pinned memory
as an upload buffer, what would the Torch code look like to wait for event
completion and upload again? Using device or stream synchronize for the
completion waiting kind of defeats the point of async copy from the CPU,
unless you can launch a bunch of other work once you started the copy in
the meantime before waiting. This becomes less viable if you have multiple
[...]

PR #223, opened 2015-08-20
Add copyAsync for asynchronous copies between host and device.

dominikgrewe:
As discussed in #216.

Only works for contiguous Float- and CudaTensors.

### DIFF
--- a/Tensor.c
+++ b/Tensor.c
@@ -51,6 +51,22 @@ static int cutorch_CudaTensor_copy(lua_State *L)
   return 1;
 }
 
+static int cutorch_CudaTensor_copyAsync(lua_State *L)
+{
+  THCState *state = cutorch_getstate(L);
+  THCudaTensor *storage = luaT_checkudata(L, 1, "torch.CudaTensor");
+  void *src;
+  if( (src = luaT_toudata(L, 2, "torch.CudaTensor")) )
+    THCudaTensor_copy(state, storage, src);
+  else if( (src = luaT_toudata(L, 2, "torch.FloatTensor")) )
+    THCudaTensor_copyAsyncFloat(state, storage, src);
+  else
+    luaL_typerror(L, 2, "torch.FloatTensor or torch.CudaTensor");
+
+  lua_settop(L, 1);
+  return 1;
+}
+
 #define CUDA_IMPLEMENT_TENSOR_COPY(TYPEC)                               \
   static int cutorch_##TYPEC##Tensor_copy(lua_State *L)                 \
   {                                                                     \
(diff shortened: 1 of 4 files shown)


dominikgrewe:
@soumith @wickedfoo Any thoughts on this?

wickedfoo:
Looks fine to me, only comment is that this won't really do much (the copy is still effectively synchronous; it copies to a small pinned staging buffer managed by the driver which is then async copied to the GPU; if the buffer is too large, it loops until the last chunk is copied over) until we have a story for CPU pinned memory allocations, and some way to prevent the GC cleanup of that memory while any copies are completing.

I think the story in the longer term to prevent GC issues would be to addref the storage while performing an async copy, then use `cudaStreamAddCallback` in the current stream for the device for which we're copying memory to/from that will add the storage to an allowed cleanup list, which we look at the next time we try to malloc or free CUDA memory, at which point we release the reference, potentially freeing the storage. The reason why we can't release the memory in the callback is because CUDA APIs apparently can't be called in the callback. I don't know in what CPU thread the callback is made either yet.

dominikgrewe:
We do have `cutorch.createCudaHostTensor` which allocates pinned memory.

You're right about the GC problem, but isn't that the case with regular CudaTensors and async operations already?

wickedfoo:
For regular CudaTensors it isn't an issue, because cudaFree is synchronizing and will wait for the completion of all operations on that memory (across all streams). I don't know if it is globally synchronizing (across all devices) or just the current device that holds the memory that is being reclaimed.

I guess the question is if freeing pinned CPU memory is synchronizing. If not, then this isn't a problem; if it is, then we need some kind of dependency wait.

wickedfoo:
(sorry edited the above comment; if it is synchronizing then it isn't an issue; if it isn't then we have to do something)

dominikgrewe:
Yes, I forgot about that. Not sure if cudaFreeHost also implicitly synchronizes. I'll try to find out.

dominikgrewe:
Thanks @wickedfoo. I've updated the patch.

wickedfoo:
looks good to me!

wickedfoo:
I guess we should figure out the story with pinned memory frees though.

soumith:
thanks @dominikgrewe

dominikgrewe:
Just to update: My contact at NVIDIA told me that `cudaFreeHost` waits for operations on _all_ devices before the memory is freed. So everything is safe.

`cudaFree` waits for operations on the device that the memory was allocated on to finish.

wickedfoo:
Ah, cool.

PR #224, opened 2015-08-20
Fix resize in addcdiv and addcmul.

### LINKED ISSUE #221: :addcdiv() and :addcmul() do not preserve shape
The :addcdiv() and :addcmul() don't preserve the existing tensor shape,
when used on a CudaTensor.

The following example shows the output size,
when a FloatTensor or a CudaTensor is used:

    torch.FloatTensor	output size:
     5
     2
    [torch.LongStorage of size 2]

    torch.CudaTensor	output size:
     10
    [torch.LongStorage of size 1]

The example:

    require 'cutorch'
    require 'cunn'

    local nY = 5
    local nX = 2
    for _, wantedType in ipairs({'torch.FloatTensor', 'torch.CudaTensor'}) do
      local input = torch.randn(nY * nX):type(wantedType)

      local output = input.new(nY, nX):zero()
      output:addcdiv(input, input)
      print(wantedType, "output size:", output:size())
    end

The nn.CDivTable() is then returning a wrong gradInput[2] shape.
I guess, `THCudaTensor_resizeAs(state, self_, src1);` should not be used
for the add*() methods.

dominikgrewe:
Fixes #221

### DIFF
--- a/lib/THC/THCTensorMath.cu
+++ b/lib/THC/THCTensorMath.cu
@@ -155,10 +155,14 @@ void THCudaTensor_addcmul(THCState *state, THCudaTensor *self_, THCudaTensor *t,
     THCudaTensor_resizeAs(state, self_, t);
     THCudaTensor_copy(state, self_, t);
   }
-  THCudaTensor_resizeAs(state, self_, src1);
+  else
+  {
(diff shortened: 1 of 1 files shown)


soumith:
thanks!

PR #225, opened 2015-08-21
[RFC] Add generic CudaTensor types

adamlerer:
This is an RFC on a patch that will add support for all the torch7 tensor types to cutorch, with an eye to adding half-precision float support as well for Maxwell in a future PR.

The new tensor types are named e.g. `torch.CudaLongTensor`, and the conversion functions are named e.g. `:cudaLong()`. The Lua and C APIs for cuda float tensors is left unchanged for backwards compatibility, so cuda float tensors are still called `CudaTensor` (and `THCudaTensor`).

Only the construction, resize, and CPU copy/type-conversion API are implemented generically so far. I'll be adding some more stuff like FFI and GPU-GPU type-conversion (and tests), but I don't expect to implement the full cutorch math API generically any time soon.

Thanks!


adamlerer:
I'm going to see if I can break this into multiple commits to make the actual changes vs. moved files more clear.

dominikgrewe:
Thanks Adam! If you could break this up a bit, that'd be great.

What do you think about having `torch.CudaFloatTensor` in addition to `torch.CudaTensor` (one being essentially an alias of the other)? It seems consistent and gives us a path towards deprecating `torch.CudaTensor` if we ever wanted to. We can only do that on the Lua side I guess, but then most of our users are probably working in Lua only.

Tensor.lua
```diff
@@ -21,55 +21,55 @@ end
...
+   cudaByte   = 'torch.CudaByteTensor',
+   cudaChar   = 'torch.CudaCharTensor',
+   cudaInt    = 'torch.CudaIntTensor',
+   cudaShort  = 'torch.CudaShortTensor',
+   cudaLong   = 'torch.CudaLongTensor'
+}
 
-local function Tensor__int(self)
-   return self:type('torch.IntTensor')
+function Tensor__converter(type)
```
fmassa: can't this function be `local` ?

adamlerer:
Hi Dominik, how do you propose I break it up? I already split out the file moves to a separate commit to make the changes clearer. Would you like me to split it into THC and cutorch components?
Also just to clarify, this is just RFC, it's not in shape to be merged yet.

Tensor.lua
adamlerer: Yep, thanks

adamlerer:
Dominik, I thought about having `torch.CudaFloatTensor` in addition to `torch.CudaTensor`. On the Lua side, you can't really alias, you'd  have to have two distinct torch types, which could both wrap the same C object; then every interfacing function would have to accept both types. I'd be worried that you'd end up with fragmentation where you could almost use the types interchangeably except for some random libraries that didn't support both "aliases" and suddenly you get confusing behavior.

What do you think? Is this how you were thinking of doing the aliasing?

hughperkins:
Cant you do simply:
```
torch.CudaTensor = torch.CudaFloatTensor
```
?

adamlerer:
Yes, you could alias in this way and even copy torch7 by adding a `cutorch.setdefaulttensortype()`, but it won't be backwards compatible because `torch.type(torch.CudaTensor())` will now be `CudaFloatTensor`.
Serialization and stuff would also have to be patched up. You could also do it the opposite way, `torch.CudaFloatTensor = torch.CudaTensor`, but that has its own problems.

dominikgrewe:
I was thinking of just doing
```lua
torch.CudaFloatTensor = torch.CudaTensor
```
Beside the issues that torch.type will always return 'CudaTensor', what other problems would there be?

I think at some point we have to break backwards-compatibility, e.g. `:cuda()` should preserve the type instead of always converting to float. We're not quite there yet, but I was hoping that a `torch.CudaFloatTensor` would make the transition easier.

lib/THC/generic/THCTensor.c
```diff
@@ -0,0 +1,784 @@
...
+}
+
+real THCTensor_(get4d)(THCState *state, const THCTensor *tensor, long x0, long x1, long x2, long x3)
+{
+  THArgCheck(tensor->nDimension == 4, 1, "tensor must have four dimensions");
+  THArgCheck((x0 >= 0) && (x0 < tensor->size[0]) && (x1 >= 0) && (x1 < tensor->size[1]) && (x2 >= 0) && (x2 < tensor->size[2]) && (x3 >= 0) && (x3 < tensor->size[3]), 2, "out of range");
+  return THCStorage_(get)(state, tensor->storage, tensor->storageOffset+x0*tensor->stride[0]+x1*tensor->stride[1]+x2*tensor->stride[2]+x3*tensor->stride[3]);
+}
+
+int THCTensor_(checkGPU)(THCState *state, unsigned int nTensors, ...)
```
dominikgrewe: checkGPU should really work with different types of CUDA tensors, but I guess there's currently no easy way to do that. Yet another reason for templated tensors :)

adamlerer: Well... there are ways you could do it. Like, if you static_assert that offsetof(data) is the same for all types, and then just casting to THCTensor does work. It's fragile though.

However, your point about templates is really my biggest question - how serious is the plan to replace this all with templates? If we're moving to templates, this effort is a waste.

adamlerer:
@andresy I'm working on this large-ish change to support non-float CudaTensor types. Before I do too much more work, can you comment on whether you're planning to switch to C++ templates in torch7 any time soon? If so, I'm just wasting my time.

adamlerer:
My latest update adds a bunch of the missing pieces - (slow) cuda->cuda copies, generic FFI, indexing, and simple tests.

It would be nice to get this reviewed now, since I would work on half-precision next. @dominikgrewe do you have any preferred way you'd like me to split this up to make it easier for review?

andresy:
@adamlerer no c++ in the short future, sorry

dominikgrewe:
I actually don't think this needs further splitting. It's a big CL but all changes are related to each other.

adamlerer:
I just got back to this PR; it's rebased on master and I've added GPU-GPU copies and tests. It's still bare-bones (no math, indexing, non-contiguous copies) but I think it's enough for one PR. @dominikgrewe would you like to take a look? Thanks!

generic/CTensor.c
```diff
@@ -0,0 +1,261 @@
...
+  if( (src = luaT_toudata(L, 2, "torch.CudaTensor")) )
+    THCTensor_(copy)(state, storage, src);
+  else if( (src = luaT_toudata(L, 2, "torch.FloatTensor")) )
+    THCTensor_(copyAsyncCPU)(state, storage, src);
+  else
+    luaL_typerror(L, 2, "torch.FloatTensor or torch.CudaTensor");
+
+  lua_settop(L, 1);
+  return 1;
+}
```
dominikgrewe: Shouldn't this be generic, rather than specifically for CudaTensor and FloatTensor?

generic/CTensor.c
```diff
@@ -0,0 +1,261 @@
...
+  THTensor *storage = luaT_checkudata(L, 1, "torch.FloatTensor");
+  void *src;
+  if( (src = luaT_toudata(L, 2, "torch.CudaTensor")) )
+    THTensor_(copyAsyncCuda)(cutorch_getstate(L), storage, src);
+  else
+    luaL_typerror(L, 2, "torch.CudaTensor");
+
+  lua_settop(L, 1);
+  return 1;
+}
```
dominikgrewe: This should also be generic

generic/CTensor.c
```diff
@@ -0,0 +1,261 @@
...
+    luaL_typerror(L, 2, "torch.*Tensor");
+
+  lua_settop(L, 1);
+  return 1;
+}
+
+static int cutorch_Tensor_(copyAsyncCPU)(lua_State *L)
+{
+  THCState *state = cutorch_getstate(L);
+  THCTensor *storage = luaT_checkudata(L, 1, "torch.CudaTensor");
```
dominikgrewe: Rename `storage` to `tensor`

generic/CTensor.c
```diff
@@ -0,0 +1,261 @@
...
+  else
+    luaL_typerror(L, 2, "torch.*Tensor");
+
+  lua_settop(L, 1);
+  return 1;
+}
+
+static int TH_CONCAT_3(cutorch_,Real,Tensor_copyAsyncCuda)(lua_State *L)
+{
+  THTensor *storage = luaT_checkudata(L, 1, "torch.FloatTensor");
```
dominikgrewe: Rename `storage` to `tensor`

generic/CTensor.c
```diff
@@ -0,0 +1,261 @@
...
+                            nElement);
+
+  THFree(d_self_sz);
+  THFree(d_self_st);
+  THFree(d_src_sz);
+  THFree(d_src_st);
+
+  lua_settop(L, 1);
+  return 1;
+}
```
dominikgrewe: Do you know why this is here? I realize it was there before, but I wonder what it's for :)

test/test.lua
```diff
@@ -2018,6 +2018,57 @@ function test.multi_gpu_copy_noncontig()
...
+      {'cudaDouble','torch.CudaDoubleTensor'},
+   }
+
+   local N = 100
+   local t0 = torch.range(1,12):reshape(3,4)
+
+   local t = t0:clone()
+   for i = 1, N do
+     -- convert to a random (CPU or GPU) type)
+     local ct = types[torch.random(#types)]
```
dominikgrewe: The test would be easier to read if you unpacked that into for example `conversionFunc` and `tensorType`

test/test.lua
```diff
@@ -2018,6 +2018,57 @@ function test.multi_gpu_copy_noncontig()
...
+     -- check the type
+     tester:assert(t:type() == ct[2], t:type() .. ' ~= ' .. ct[2])
+
+     -- check metadata
+     tester:assert(t:isContiguous())
+     tester:assert(t:size(1) == 3 and t:size(2) == 4)
+     tester:assert(t:nDimension() == 2)
+
+     -- check data
+     tester:assert((t:double()-t0):abs():max() == 0)
```
dominikgrewe: use `tester:assertTensorEq(t:double(), t0, 0)`

test/test.lua
```diff
@@ -2018,6 +2018,57 @@ function test.multi_gpu_copy_noncontig()
...
+      {'cudaShort', 'torch.CudaShortTensor'},
+      {'cudaLong',  'torch.CudaLongTensor'},
+      {'cudaDouble','torch.CudaDoubleTensor'},
+   }
+
+   local N = 100
+   local t0 = torch.range(1,12):reshape(3,4)
+
+   local t = t0:clone()
+   for i = 1, N do
```
dominikgrewe: Why do you loop N times and pick a random type instead of simply iterating over all types?

test/test.lua
```diff
@@ -2018,6 +2018,57 @@ function test.multi_gpu_copy_noncontig()
...
+
+     -- check metadata
+     tester:assert(t:isContiguous())
+     tester:assert(t:size(1) == 3 and t:size(2) == 4)
+     tester:assert(t:nDimension() == 2)
+
+     -- check data
+     tester:assert((t:double()-t0):abs():max() == 0)
+
+
```
dominikgrewe: Check `copy` as well as type conversion?

generic/CStorage.c
```diff
@@ -0,0 +1,122 @@
...
+  else if( (src = luaT_toudata(L, 2, "torch.LongStorage")) )
+    THStorage_(copyLong)(storage, src);
+  else if( (src = luaT_toudata(L, 2, "torch.FloatStorage")) )
+    THStorage_(copyFloat)(storage, src);
+  else if( (src = luaT_toudata(L, 2, "torch.DoubleStorage")) )
+    THStorage_(copyDouble)(storage, src);
+  else if( (src = luaT_toudata(L, 2, "torch.CudaStorage")) )
+    THStorage_(copyCudaFloat)(cutorch_getstate(L), storage, src);
+  else if( (src = luaT_toudata(L, 2, "torch.CudaLongStorage")) )
+    THStorage_(copyCudaLong)(cutorch_getstate(L), storage, src);
```
dominikgrewe: Why only Float and Long?

generic/CTensor.c
adamlerer: This just got added... I wasn't sure it was really necessary but I can add it.

generic/CTensor.c
adamlerer: The trail peters out at cuda_FloatTensor_fakecopy. @wickedfoo ?

test/test.lua
adamlerer: Because there are n^2 conversions from each type to each other type, so this gets better converage.

generic/CStorage.c
adamlerer: oops. Should add a test for storage as well.

test/test.lua
dominikgrewe: Oh, I see. The iterations aren't independent but the `t` carries over from one iteration to the next. Might be worth stating that :)

adamlerer:
Updated with changes suggested by @dominikgrewe

dominikgrewe:
I don't think I can see any changes.

adamlerer:
Sorry @dominikgrewe, got git and hg mixed up again :/. Files should be updated now.

test/test.lua
```diff
@@ -2018,6 +2018,110 @@ function test.multi_gpu_copy_noncontig()
...
+   local t0 = torch.range(1,12):reshape(3,4)
+
+   -- t carries over from one iteration to the next
+   local t = t0:clone()
+   for i = 1, N do
+      -- convert to a random (CPU or GPU) type)
+      local ct = types[torch.random(#types)] 
+      local conversionFunc = ct[1]
+      local tensorSubtype = ct[2]
+      local tensorType = 'torch.' .. ct[2]
```
dominikgrewe: You can just do this:
```lua
local conversionFunc, tensorSubtype = unpack(types[torch.random(#types)])
```

test/test.lua
```diff
@@ -2018,6 +2018,110 @@ function test.multi_gpu_copy_noncontig()
...
+   local N = 100
+   local t0 = torch.range(1,12):reshape(3,4):storage()
+
+   -- t carries over from one iteration to the next
+   local t = torch.DoubleStorage(t0:size()):copy(t0)
+   for i = 1, N do
+      -- convert to a random (CPU or GPU) type)
+      local ct = types[torch.random(#types)] 
+      local conversionFunc = ct[1]
+      local storageSubtype = ct[2]
```
dominikgrewe: same here, just use `unpack`

dominikgrewe:
Just two nits, otherwise it looks good to me!

adamlerer:
Nits were fixed, lets merge this soon to avoid further rebasing work?

soumith:
Ready to merge on rebase. Theres merge conflicts.

adamlerer:
Fixed!

soumith:
Kaboom! Done!

Issue #230, opened 2015-09-01
scatter has missing assertions


soumith:
Reported by @lvdmaaten cc: @dominikgrewe 

There are assertions missing on the first input of :scatter() in the GPU implementation. This may lead to illegal memory accesses.

Repro:

```lua
local ac = torch.FloatTensor(16, 1000):fill(-1)
local ag = torch.CudaTensor(16, 1000):fill(-1)
local ind = torch.FloatTensor{
    {198},
    { 56},
    {939},
    {906},
    {622},
    {712},
    {438},
    {446},
    {702},
    {155},
    {298},
    {366},
    {784},
    {834},
    {616},
    {238}
}  -- 16x1 Tensor with values >=1 and <=1000

-- erroneously providing 1 as first input instead of 2:
ag:scatter(1, ind:long(), 1)  -- no assertion here: leads to illegal memory access
ac:scatter(1, ind:cuda(), 1)  -- an assertion catches this
```

albanD:
@soumith I think there is a small typo in the two last lines of your sample:
```lua
ag:scatter(1, ind:cuda(), 1)  -- no assertion here: leads to illegal memory access
ac:scatter(1, ind:long(), 1)  -- an assertion catches this
```

lvdmaaten:
I don't think there is a typo? Both commands should fail because the first input should be 2 instead of 1. However, the GPU implementation (first line) does not appropriately check the inputs and so it actually executes the kernel, which as a result writes in memory where it should not be writing. The solution is to add the appropriate assertions to the GPU implementation of scatter() to prevent this from happening.

albanD:
I agree that both line should fail. And the missing assert in the cuda side leads to illegal memory access.

My remark was just about the sample code provided, the first line where you test for the cuda implementation, indices should be cast to cuda (otherwise you have an unexpected argument error).
In the second line, the indices should be cast to long (otherwise you have an unexpected argument error).
As I said its just a typo in the sample code: the commands fails for the wrong reason (incorrect type of the indice tensors). If you switch as I proposed, the fail for the reason that is interesting for us (one for illegal access, the other for inconsistent tensor size)

lvdmaaten:
Oh I see... sorry, I misread that!

PR #235, opened 2015-09-08
Add missing size check to scatterFill.

dominikgrewe:
Checks that the index tensor has the right shape.


dominikgrewe:
Fixes #230

soumith:
thanks Dominik

Issue #246, opened 2015-09-29
indexSelect1 test flaky


dominikgrewe:
The test fails for certain random seeds, e.g. 1443533672:

```bash
seed: 	1443533672	
Running 1 tests	
*  ==> Done        

Completed 0 asserts in 1 tests with 1 errors	

--------------------------------------------------------------------------------	
indexSelect1
 Function call failed 
/usr/share/lua/5.1/cutorch/test.lua:71: bad argument #3 to 'fn' (out of range at /usr/local/google/code/sdb1/deepmind/torch/torch7/lib/TH/generic/THTensor.c:370)
stack traceback:
	[C]: in function 'fn'
	/usr/share/lua/5.1/cutorch/test.lua:71: in function 'compareFloatAndCuda'
	/usr/share/lua/5.1/cutorch/test.lua:1098: in function 'v'
	/usr/share/lua/5.1/cutorch/test.lua:2522: in function </usr/share/lua/5.1/cutorch/test.lua:2520>
	[C]: in function 'xpcall'
	/usr/share/lua/5.1/torch/Tester.lua:115: in function 'pcall'
	/usr/share/lua/5.1/torch/Tester.lua:186: in function '_run'
	/usr/share/lua/5.1/torch/Tester.lua:161: in function 'run'
	/usr/share/lua/5.1/cutorch/test.lua:2539: in function 'test'
	(command line):1: in main chunk
	[C]: at 0x00405e60
	
--------------------------------------------------------------------------------	

```

hughperkins:
Observation: solvable by adding `cltorch.manualSeed(123)` at start of test.  Repeatable, deterministic.

dominikgrewe:
Or just pass the random seed on the command line:
```bash
th -lcutorch -e "cutorch.test({'indexSelect1'}, 1443533672)"
```

dominikgrewe:
@oyamauchi Seems like this started after your recent PR. Not sure if this was introduced by your changes or just surfaced by the new test.

hughperkins:
Yes ... but out of the box, when users run these, they presumably are not going to do that, and the tests will fail randomly.  I'd rather that the tests run in a loop a few times if we feel the need, but the set of looped tests will either always pass, or always fail (the seed will be before the loop, so each iteration through the loop will get different numbers).

dominikgrewe:
[...]

PR #247, opened 2015-09-29
Fix indexSelect1 test.

dominikgrewe:
Pick indices after transposing tensor.


dominikgrewe:
Fixes #246

soumith:
thanks for the quick fix!

Issue #171, opened 2015-06-04
Rerunning make causes whole cutorch to rebuild again


hughperkins:
modifying one tiny file (or even, not modifying anything...), and then rerunning `luarocks make rocks/cutorch-scm-1.rockspec` causes every file to be rebuilt again, which takes about 5 minutes each time.

I suspect this might be connected to https://github.com/torch/cutorch/issues/100 ?

hughperkins:
Seems like this is something built into cmake's FindCUDA module perhaps?  There seems nothing in the cutorch lib/THC/CMakeLists.txt about this, and the error messages all mention /usr/share/cmake-2.8/Modules/FindCUDA.  (I'm waiting for cutorch to build again, hence have time to ponder this issue :-P )

szagoruyko:
CUDA_ADD_LIBRARY rebuilds everything on any make call, to my experience

hughperkins:
So, I had to rebuild cunn a few times this afternoon, so I hacked around a bit, to see if can prevent the whole-world-building experience, and had some success :-)  Using root, open `/usr/share/cmake-2.8/Modules/FindCUDA.cmake`, and scroll down to line 1295 approx, and modify the start of the sectoin there as follows:

```
      # Build the generated file and dependency file ##########################
      message("generated_file ${generated_file}")
      message("CUDA_NVCC_DEPEND ${CUDA_NVCC_DEPEND}")
      message("custom_target_script ${custom_target_script}")
      message("CMAKE_BINARY_DIR ${CMAKE_BINARY_DIR}")
      set(newdepend "")
      foreach(depend ${CUDA_NVCC_DEPEND})
          if(NOT ${depend} STREQUAL "${CMAKE_CURRENT_BINARY_DIR}/CMakeFiles/${cuda_target}.dir")
              set(newdepend "${newdepend};${depend}")
          endif()
      endforeach()
      set(CUDA_NVCC_DEPEND "${newdepend}")
      add_custom_command(
        OUTPUT ${generated_file}
        # These output files depend on the source_file and the contents of cmake_dependency_file
        #${main_dep}
        DEPENDS ${CUDA_NVCC_DEPEND}
        #DEPENDS ${custom_target_script}
```

Now, if you build the first time, it of course has to build everyhting.  Bu tthe next time, it will just build hte files that have chnaged :-)

Tested using cunn.

hughperkins:
This is fixed by installing https://github.com/hughperkins/FindCUDA

Issue #100, opened 2015-01-13
Annoying generic


szagoruyko:

hughperkins:
I think the dependencies do exist, just not being located.  I tried adding the `TH` path to include directories, and that's not sufficient, but I reckon something along those lines might somehow fix it.

```
INCLUDE_DIRECTORIES("${Torch_INSTALL_INCLUDE}/TH")
```

hughperkins:
Rather than commenting out in make2cmake, one can comment out the following in `/usr/share/cmake-2.8/Modules/FindCUDA/run_nvcc.cmake` instead.  There might be consequences to this, but doing `touch` on a .cu file seems to trigger rebuild, when calling via `luarocks make`.

```
# Generate the dependency file
#cuda_execute_process(
#  "Generating dependency file: ${NVCC_generated_dependency_file}"
#  COMMAND "${CUDA_NVCC_EXECUTABLE}"
#  -M
#  ${CUDACC_DEFINE}
#  "${source_file}"
#  -o "${NVCC_generated_dependency_file}"
#  ${CCBIN}
#  ${nvcc_flags}
#  ${nvcc_host_compiler_flags}
#  ${depends_CUDA_NVCC_FLAGS}
#  -DNVCC
#  ${CUDA_NVCC_INCLUDE_ARGS}
#  )

#if(CUDA_result)
#  message(FATAL_ERROR "Error generating ${generated_file}")
#endif()

# Generate the cmake readable dependency file to a temp file.  Don't put the
# quotes just around the filenames for the input_file and output_file variables.
# CMake will pass the quotes through and not be able to find the file.
#cuda_execute_process(
#  "Generating temporary cmake readable file: ${cmake_dependency_file}.tmp"
#  COMMAND "${CMAKE_COMMAND}"
#  -D "input_file:FILEPATH=${NVCC_generated_dependency_file}"
#  -D "output_file:FILEPATH=${cmake_dependency_file}.tmp"
#  -P "${CUDA_make2cmake}"
#  )

#if(CUDA_result)
#  message(FATAL_ERROR "Error generating ${generated_file}")
#endif()

# Copy the file if it is different
#cuda_execute_process(
#  "Copy if different ${cmake_dependency_file}.tmp to ${cmake_dependency_file}"
#  COMMAND "${CMAKE_COMMAND}" -E copy_if_different "${cmake_dependency_file}.tmp" "${cmake_dependency_file}"
#  )

#if(CUDA_result)
[...]

PR #251, opened 2015-10-03
EXPERIMENTAL improve dependencies during build

hughperkins:
This is mostly for discussion purposes for now.  In theory, what it does is:
- if you build twice, without making any changes, the second time, bulid almost instant
- if you then `touch` one .cu file, and rebuild, that .cu file, and only that .cu file, will be rebuilt
- no more messages about 'dependency not found, removing'

Quite experimental for now, not very tested.  I don't reckon we should merge to master for now, but maybe worth trying for a bit, see how it goes?

Relates to following issues:
- [Annoying non-existent dependency file warnings](https://github.com/torch/cutorch/issues/100)
- [Rerunning make causes whole cutorch to rebuild again](https://github.com/torch/cutorch/issues/171)


Atcold:
> no more messages about 'dependency not found, removing'

Sweet!

hughperkins:
Thanks!   I reckon it'd be good for a few people to try it first though, and check does/doesnt work.  Actually, I think I might split into two commits, so it's clearer what it does, which will be:
- commit 1: copy FindCUDA files into cutorch, renaming them to FindCUDA_streamlined, to avoid naming conflicts
- commit 2: modification to how dependency handling works

hughperkins:
How this is intended to work:

Background:
- dependency handling is handled by the `add_custom_command` at https://github.com/hughperkins/cutorch/blob/a9dc0709c900295ba2731d18083d7800e815d880/cmake/FindCUDA_streamlined.cmake#L1294-L1296
  - it has a looott of dependencies, including .cu files (good), .h files (good), the cunn.dir directory (less good), the generated build script itself (less good)
- it's handled all over again during build by the lines at https://github.com/hughperkins/cutorch/blob/a9dc0709c900295ba2731d18083d7800e815d880/cmake/FindCUDA_streamlined/run_nvcc.cmake#L181-L234
  - these are superfluous I think, since we're always running cmake, via luarocks
  - these lines rerun `nvcc -M`, to get dependencies list
  - these lines are the ones created the `dependency removed` messages

So, in this change:
- remove some of the dependencies from the `add_custom_command` command, eg the dependnecy on the build script, and on `cunn.dir`
- remove the superfluous additional checking from `run_nvcc.cmake`

hughperkins:
(hmmm, but the new, unsquashed, version doesnt seem to fix the dependencies :-P  checking...)

hughperkins:
Hmmm, seems not to work now :-O  Maybe I still had some hacks in my global directories, and after reinstalling `cmake-data`, the contents of CUDA_NVCC_DEPEND at line 1298 seem radically different :-O

hughperkins:
(Closing this for now, since doesnt work at moment :-P )

hughperkins:
Hmmm, cant reopen after force pushing :-(

PR #253, opened 2015-10-05
Implementation of torch.cat

dominikgrewe:


dominikgrewe:
This is pretty much copy-and-paste from Torch.

soumith:
thank you!

Issue #239, opened 2015-09-13
cutorch.test() fails on multi_gpu_copy if GPUs don't support P2P


skelleher:
cutorch.test() crashes on multi_gpu_copy_noncontig if there are GPUs which aren't identical / do not support peer access.  

If I limit the visible GPUs to 1 the test does not crash.

[cooper@titan]: export CUDA_VISIBLE_DEVICES=0,1                           
[cooper@titan]: th -lcutorch -e "cutorch.test{'multi_gpu_copy_noncontig'}"
seed: 	1442182014	
Running 1 tests	
|  ==> multi_gpu_copy_noncontigcuda runtime error (77) : an illegal memory access was encountered at /tmp/luarocks_cutorch-scm-1-4103/cutorch/lib/THC/THCStorage.c:122	

[cooper@titan]: export CUDA_VISIBLE_DEVICES=0                             
[cooper@titan]: th -lcutorch -e "cutorch.test{'multi_gpu_copy_noncontig'}"
seed: 	1442182030	
Running 1 tests	
_  ==> Done Completed 4 asserts in 1 tests with 0 errors	
--------------------------------------------------------------------------------

soumith:
cc: @wickedfoo wrote patches for this, he has to push them upstream.

soumith:
the patches are to enable non-p2p copies in general, when p2p isn't available..

soumith:
should be fixed now i think.

Issue #212, opened 2015-07-29
Cross-device tensor copy has inconsistent/weird/broken stream behavior


wickedfoo:
Which stream the cross-device copy uses has has different rules at present depending upon whether it is contig or non-contig, which is rather non-obvious to the user. This causes problems when you attempt to use non-default streams, which I'm getting very deep into at the moment.

The code:
```
  bool succ =
        THCudaTensor_pointwiseApply2(state, dst, src, CopyOp<float>());
```

can actually not use a stream on either the src or dest device (just whatever the current device is), since pointwiseApply is getting its stream via THCState, whereas that is not changed when you call cudaSetDevice.

It is unclear to me if cudaMemcpyAsync allows you to use a stream that has nothing to do with the src or dst devices. Right now, you can cutorch.setDevice() to a device that has nothing to do with the src or dst devices, and it's kind of up in the air as to what that means. Perhaps that should be an error.

I propose changing the tensor copy code to obey the following rules:

===
1. All copies (contig or not) are made from the src tensor device (regardless of what cutorch.getDevice returns at the point of :copy() invocation), using whatever stream the current stream index is on the src device. This removes the "optimization" for non-contig copies  but this makes it more sensible I think, so you always know what stream/device the copy is ordered with respect to.
Looking at nvprof/nvvp, if you issue the copy from a stream off of the src device, the CUDA API creates an internal stream for use on the src device, and adds additional synchronization. If you issue the copy from a src device stream, the p2p copy always occurs on the stream you intend.

2. If a copy is made on the default stream (THCState_getCurrentStreamIndex(state) == 0, regardless of what the current cutorch.getDevice returns at the point of :copy() invocation), then the default stream on both the src and dst devices will have a full barrier:

  -src stream 0 syncs to dst stream 0 (no one should be writing into dst before starting)
  -copy occurs on src device, stream 0
  -dst stream 0 syncs to src stream 0 (the copy should be complete before dst continues)

[...]

Issue #134, opened 2015-04-10
Non-contiguous copy fails between two GPUs without p2p access


dominikgrewe:
Assuming GPUs 1 and 2 don't support peer-to-peer access, the following code fails:
```lua
require 'cutorch'

x = torch.CudaTensor(10, 10):t()
cutorch.setDevice(2)
y = torch.CudaTensor(10, 10):t()

print('copy')
x:copy(y)
print('sync')
cutorch.synchronize()
print('done')
```

shuzi:
@dominikgrewe 

Hello,

is your problem solved?
I run your code, it print out 'copy' and 'sync' and then the program is dead.
the problem seem to be cutorch.synchronize()  ?

hughperkins:
@shuzi: Per my understanding, a broken kernel, eg out-of-bounds array access, will cause the commandqueue/stream to crash, but you wont actually notice until you call finish/synchronize.  So, you might think that the problem is in 'sync', but that's just because sync is where it checks for commandqueue errors.

wickedfoo:
this should be fixed now with the above diff.

PR #254, opened 2015-10-05
cutorch copy fixes, add CUDA Event API to Lua

wickedfoo:
Fixes:

https://github.com/torch/cutorch/issues/239
https://github.com/torch/cutorch/issues/212
https://github.com/torch/cutorch/issues/134

If p2p access cannot be enabled, then we default to make contig + cudaMemcpyAsync. This just happens for all GPUs instead of trying to restrict it to the pairs of GPUs for which p2p access cannot be enabled.

Implements the async copy behavior changes outlined in #212.

Also adds a Lua-level CUDA event API. GC of this object will free the CUDA event.

Also makes cutorch mirror the `cutorch` table definition that torch7 has, a change @andresy wanted me to make when I was mucking with the event stuff internally.

### DIFF
--- a/init.c
+++ b/init.c
@@ -16,7 +16,7 @@ int checkAndCountListOfStreams(lua_State *L, THCState *state, int arg,
                                int device)
 {
   if (!lua_istable(L, arg)) {
-    THError("expecting table of device streams");
+    THError("expecting array of device streams");
   }
 
   /* Push table to top */
@@ -26,8 +26,11 @@ int checkAndCountListOfStreams(lua_State *L, THCState *state, int arg,
   int streams = 0;
   lua_pushnil(L);
   while (lua_next(L, -2)) {
+    if (!lua_isnumber(L, -2)) {
+      THError("expected array of streams, not table");
+    }
     if (!lua_isnumber(L, -1)) {
-      THError("streamWaitFor: list of streams must be numeric");
+      THError("array of stream ids must contain numeric ids");
     }
     int streamId = (int) lua_tonumber(L, -1);
 
@@ -117,7 +120,7 @@ void createMultiDeviceEvents(lua_State *L, THCState *state, int arg,
   while (lua_next(L, -2)) {
     int device = (int) lua_tonumber(L, -2) - 1;
     THCudaCheck(cudaSetDevice(device));
-    events += createSingleDeviceEvents(L, state, -1, device, events);    
+    events += createSingleDeviceEvents(L, state, -1, device, events);
     ++gpu;
 
     lua_pop(L, 1);
@@ -138,7 +141,7 @@ void waitSingleDeviceEvents(lua_State *L, THCState *state, int arg,
   lua_pushnil(L);
   while (lua_next(L, -2)) {
     int streamId = (int) lua_tonumber(L, -1);
-    cudaStream_t stream = 
+    cudaStream_t stream =
       THCState_getDeviceStream(state, device, streamId);
     for (int i = 0; i < numEvents; i++) {
       THCudaCheck(cudaStreamWaitEvent(stream, event[i], 0));
@@ -374,7 +377,7 @@ static int cutorch_streamWaitFor(lua_State *L)
   }
   /* One-way dependency; streamWaiting will wait for the list of streams to
      wait on to complete execution of pending scheduled kernels/events */
-  cudaEvent_t * events = (cudaEvent_t*)malloc(sizeof(cudaEvent_t) * streams);  
+  cudaEvent_t * events = (cudaEvent_t*)malloc(sizeof(cudaEvent_t) * streams);
   createSingleDeviceEvents(L, state, 2, curDev, events);
   /* Then, wait on them */
   for (int i = 0; i < streams; i++) {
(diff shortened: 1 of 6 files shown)


init.c
```diff
@@ -16,7 +16,7 @@ int checkAndCountListOfStreams(lua_State *L, THCState *state, int arg,
                                int device)
 {
   if (!lua_istable(L, arg)) {
-    THError("expecting table of device streams");
+    THError("expecting array of device streams");
```
wickedfoo: also FYI this stream/device wait API could now be fully implemented in Lua if we want to using cutorch.Event().
these were some error message changes I made a while ago. really I should have made this API use a table whose keys were streams instead to enforce uniqueness.

wickedfoo:
ping?

init.c
```diff
@@ -785,6 +830,8 @@ LUA_EXTERNC DLL_EXPORT int luaopen_libcutorch(lua_State *L);
 int luaopen_libcutorch(lua_State *L)
 {
   lua_newtable(L);
+  lua_pushvalue(L, -1);
+  lua_setglobal(L, "cutorch");
```
dominikgrewe: Why are you making this global? init.lua already creates a global "cutorch" table. And in fact we want to avoid creating globals altogether (see [Torch Roadmap](https://github.com/torch/torch7/blob/master/ROADMAP.md#versioning--packaging))

dominikgrewe: Oh, just saw you changed that in init.lua. Is there a particular reason for that?

lib/THC/THCGeneral.c
```diff
@@ -136,24 +142,38 @@ void THCudaEnablePeerToPeerAccess(THCState* state)
...
+          /* of p2p access. */
+          state->peerToPeerAccessEnabled = 0;
+          break;
         }
       }
     }
+
+    if (!state->peerToPeerAccessEnabled) {
+      break;
+    }
```
dominikgrewe: I don't like this change. What if you have clusters of GPUs within a machine where GPUs have peer access within a cluster but not across clusters? You still want to use p2p within a cluster. If you're worried about cudaDeviceCanAccessPeer being slow, why don't we create a look-up table ourselves?

lib/THC/THCTensorCopy.cu
```diff
@@ -28,62 +28,122 @@ THCudaTensor_copy(THCState* state, THCudaTensor* dst, THCudaTensor* src) {
...
+    // If both src and dst are on the same device, or if they are on
+    // different devices and p2p access is enabled, perform the copy
+    // by a pointwise copy kernel.
+    // Otherwise, we'll have to make contiguous (which will in fact
+    // invoke copy() again), and then perform the copy.
+    // FIXME: might want to consider only running the pointwise kernel
+    // if both src and dst innermost dimensions are contiguous. If
+    // they are not, then taking the hit of the memory allocation/free
+    // might be worth it to avoid non-coalesced reads or writes.
+    if ((srcDev == oldDev) ||
```
dominikgrewe: Should this be `srcDev == dstDev`?

lib/THC/THCTensorCopy.cu
```diff
@@ -28,62 +28,122 @@ THCudaTensor_copy(THCState* state, THCudaTensor* dst, THCudaTensor* src) {
...
-    THCudaCheck(cudaEventDestroy(doneCopying));
+      // Restore prior THCState stream
+      if (srcDev != oldDev) {
+        THCState_setStream(state, oldDev, copyStreamIndex);
+      }
+    } else {
+      // GPUs can't access each other directly; fall back to
+      // newContiguous and memcpy
+      THCudaTensor* srcContig = THCudaTensor_newContiguous(state, src);
+      THCudaTensor* dstContig = THCudaTensor_newContiguous(state, dst);
```
dominikgrewe: If `dst` in non-contiguous this creates an unnecessary copy.

lib/THC/THCTensorCopy.cu
```diff
@@ -28,62 +28,122 @@ THCudaTensor_copy(THCState* state, THCudaTensor* dst, THCudaTensor* src) {
...
+
+      THCudaCheck(cudaMemcpyAsync(THCudaTensor_data(state, dstContig),
+                                  THCudaTensor_data(state, srcContig),
+                                  totalElements * sizeof(float),
+                                  cudaMemcpyDeviceToDevice,
+                                  copyStream));
+
+      THCudaTensor_free(state, srcContig);
+      THCudaTensor_set(state, dst, dstContig);
```
dominikgrewe: I don't think `set` is the right thing to do. If `dst` shares its storage with another tensor, you expect this storage to be modified by the copy. So I think you have to do a copy here. `freeCopyTo` is probably the right function which also cleans up the memory of `dstContig`.

init.c
wickedfoo: @andresy wanted me to make that change

lib/THC/THCTensorCopy.cu
wickedfoo: the copy is unavoidable, without p2p you can't copy using a kernel if dst is non-contig, you have to use cudaMemcpyAsync which will route the copy via the host. if dst is contig then it just addrefs it.

in theory one could see if both are non-contiguous in the same manner and if there are no holes and one can do the copy, but that might be an excessive check

lib/THC/THCTensorCopy.cu
wickedfoo: this one yes, the ones below no. will fix

init.c
wickedfoo: and torch7 does it the same way

lib/THC/THCGeneral.c
wickedfoo: sure, i'll create a map and use that.

lib/THC/THCTensorCopy.cu
wickedfoo: sounds good, though that's another potential copy, though i think we do in fact use the _freeCopyTo pattern elsewhere too.

wickedfoo:
sorry i only saw your comments now, i'll update

lib/THC/THCTensorCopy.cu
dominikgrewe: I just thought it was unnecessary to actually copy the data in the output tensor, because we immediately overwrite it. It would be more efficient to simply create a new tensor of the right size (without copying any data), if it is non-contiguous.

lib/THC/THCTensorCopy.cu
dominikgrewe: Yes, we use it in quite a few places for the reasons I mentioned above.

lib/THC/THCTensorCopy.cu
wickedfoo: ah i see, will do that.

wickedfoo:
one min, forgot cleanup

wickedfoo:
ok should be good to go, now integrated cross-device / disabled p2p access in the randomized copy test, which hits the non-contig pathways too.

i maintain the matrix of (device i -> can access device j via p2p) as well, and use that for the copy. this can be selectively disabled or re-enabled as well.

dominikgrewe:
Thanks Jeff!

PR #255, opened 2015-10-05
faster norm/normall code

wickedfoo:
Use template specialization to avoid unnecessary pow, switch to float form.

Speedup for normall (2-norm):

old code 

tensor size | time (ms)
----|----
10000 | 0.19192695617676
100000 | 0.18906593322754
1000000 | 0.31685829162598
10000000 | 1.3060569763184
100000000 | 11.244058609009

new code 

tensor size | time (ms)
----|----
10000 | 0.12993812561035
100000 | 0.15497207641602
1000000 | 0.14400482177734
10000000 | 0.34999847412109
100000000 | 2.2017955780029

### DIFF
--- a/lib/THC/THCTensorMath2.cu
+++ b/lib/THC/THCTensorMath2.cu
@@ -426,24 +426,28 @@ void THCudaTensor_std(THCState *state, THCudaTensor *self_, THCudaTensor *src, l
   THCudaTensor_freeCopyTo(state, self, self_);
 }
 
-
-struct norm_functor
+template <int StaticExp>
+struct TensorNormOp
 {
-  const float exponent;
+  TensorNormOp(float exp) : exponent(exp) {}
 
-  norm_functor(float exponent_) : exponent(exponent_) {}
-
-    __host__ __device__ float operator()(const float& x) const
-  {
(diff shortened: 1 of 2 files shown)


lib/THC/THCTensorMath2.cu
```diff
@@ -426,24 +426,30 @@ void THCudaTensor_std(THCState *state, THCudaTensor *self_, THCudaTensor *src, l
...
 
-  norm_functor(float exponent_) : exponent(exponent_) {}
+  __host__ __device__ float operator()(float x) const {
+    float val = fabsf(x);
 
-    __host__ __device__ float operator()(const float& x) const
-  {
-    return pow(fabs(x), exponent);
+    if (StaticExp == 1) {
+      return x;
```
fmassa: shouldn't it return `val` instead of `x` ?

wickedfoo: ah yes, will fix the test to catch this too.

wickedfoo:
fixed now, and test checks for that with negative numbers (fails without the change)

dominikgrewe:
Thanks!

PR #256, opened 2015-10-06
EXPERIMENTAL improve dependencies during build

hughperkins:
This is mostly for discussion purposes for now.  In theory, what it does is:
- if you build twice, without making any changes, the second time, bulid almost instant
- if you then `touch` one .cu file, and rebuild, that .cu file, and only that .cu file, will be rebuilt
- no more messages about 'dependency not found, removing'

Quite experimental for now, not very tested.  I don't reckon we should merge to master for now, but maybe worth trying for a bit, see how it goes?

Relates to following issues:
- [Annoying non-existent dependency file warnings](https://github.com/torch/cutorch/issues/100)
- [Rerunning make causes whole cutorch to rebuild again](https://github.com/torch/cutorch/issues/171)

How this is intended to work:

Background:
- dependencies are added to `add_custom_command` at https://github.com/hughperkins/cutorch/blob/a9dc0709c900295ba2731d18083d7800e815d880/cmake/FindCUDA_streamlined.cmake#L1294-L1296
  - it has a looott of dependencies, including .cu files (good), .h files (good), the cunn.dir directory (less good), the generated build script itself (less good)
  - first time through, it simply depends on `.depend` file
  - second time you run, it actually loads all the dependencies from that file, into CUDA_NVCC_DEPEND, and we can filter that a bit, if we want
- dependency file generation is handled by the lines at https://github.com/hughperkins/cutorch/blob/a9dc0709c900295ba2731d18083d7800e815d880/cmake/FindCUDA_streamlined/run_nvcc.cmake#L181-L234
  - these lines run `nvcc -M`, to get dependencies list
  - they trigger the `dependency removed` messages

So, in this change:
- filter out some of the dependencies from the `add_custom_command` command, eg the dependnecy on the build script, and on `cunn.dir`
- as per the change that has traditionally been applied to the global /usr/share/cmake FindCUDA module, comment out the `dependency removed` warning

This PR was originally created at https://github.com/torch/cutorch/pull/251 , but apparently cannot reopen a PR after force push, so recreating.

Edit: Example build outputs:
- example output during rebuild https://gist.github.com/hughperkins/76dcd2207892040ed711
- example output after `touch`ing THCStorageCopy.cu: https://gist.github.com/hughperkins/ad2e74661507c6797944


hughperkins:
(Hmmm, I think I shall rename from `CUDA_streamlined` to `CUDA_local`, seems more objective.

hughperkins:
Updated:
- split into 3 commits:
  - initial copy of FindCUDA, no renaming
  - rename FindCUDA to FindCUDA_local
  - improve dependnecies handling, remove warning messages
- renamed from FindCUDA_streamlined to FindCUDA_local

hughperkins:
Thoughts on this?  It's kind of frustrating changing one tiny thing, an dthe whole world rebuilds :P This PR fixes this.  Not sure how to test it though?  Seems that travis passes.  Anything else we should test?  Something I can add to travis to improve a bit?

hughperkins:
Oh wait, there is no travis right? Hmmm

soumith:
@hughperkins i'll be happy to review and merge it. However, can you generate a diff of the changes you made on top of the originals (you can put it on gist if you wish)? that'll help me with both code review and maintenance.

hughperkins:
Yes, so, the change is in three commits, so it's easy to view:
- add the original files, from nvidia https://github.com/hughperkins/cutorch/commit/0f2ecbf876b759c4469f57d82810efabcb3e29c1
- rename FindCUDA to FindCUDA_local https://github.com/hughperkins/cutorch/commit/cd7498f06e731cc9a28adee4c9b999c8d14b1578   This is so no naming conflicts with the system-installed one
- do the customizations to our local FindCUDA_local https://github.com/hughperkins/cutorch/commit/7a88102c39d5da41b3df70cea6ce4af20370c6f3

soumith:
okay. everything looks simple and good, except that I dont think it's necessary to change the name to _local. You can keep it as FindCUDA

hughperkins:
Ok. Will update, and repush.

szagoruyko:
I don't think fixing a warning is worth adding 2k lines of code that are shipped with cmake anyway. are we going to update these files everytime they are changed in cmake?

soumith:
@szagoruyko it's not just fixing the warning. this does not rebuild all cutorch files everytime we run luarocks make rocks/* , it only rebuilds files that have changed. that is a huge boost to developer productivity for us working on cutorch/cunn

szagoruyko:
oh sorry, get it now. still my question remains, it may cause problems with different versions on cmake and CUDA. we should then make it a flag option for developers and keep the normal cmake method to be used by default.

hughperkins:
I'm ok with making it optional.  As long as I dont have to wait 20 minutes every time I add a semicolon or something, I'm happy :-)  I think this would mean we would need to keep the names separate, and the switch would basically change line 7 of CMakeLists.txt, ie `FIND_PACKAGE(CUDA_local 5.5 REQUIRED)` vs `FIND_PACKAGE(CUDA 5.5 REQUIRED)`

soumith:
it's true that it'll be a maintenance headache. we used to have our own files in the past, and had lots of OSX issues. Ok I give in, we revert back to CUDA_local and propagate a flag to activate this path (maybe via an environment variable like CUTORCH_CMAKE=1 )

hughperkins:
> maybe via an environment variable like CUTORCH_CMAKE=1

Ah, because it has to propogate through the luarocks, and the rockspec?  Ok, sounds good.

borisfom:
We need this fix  badly! _local is fine with me - we should submit the fix to cmake developers though.

hughperkins:
> we should submit the fix to cmake developers though

Good idea.  I can handle that.  Let's test it a bit first though :-)  I shall change it to be an option.

hughperkins:
(ignore those commits just now; just a rebase)

hughperkins:
Ok. This last commit changes the use of local FindCUDA to an option.  Use like this:

```
CUTORCH_INCREMENTAL_BUILDS=ON luarocks make rocks/cutorch-scm-1.rockspec
```

Tests carried out:
- erase build directory
- start building using `luarocks make rocks/cutorch-scm-1.rockspec`
  - check starts building ok
  - check uses the old build process, ie with the dependency warnings etc
  - abort after 30-60 seconds of building
- start building using `CUTORCH_INCREMENTAL_BUILDS=ON luarocks make rocks/cutorch-scm-1.rockspec`
  - check warning messages disappeared
  - use `find build -name '*.o'`occasionally to check `*.o` files start appearing
  - after several `.o` files built, interrupt the build
- start build again, using incremental options, ie `CUTORCH_INCREMENTAL_BUILDS=ON luarocks make rocks/cutorch-scm-1.rockspec`
  - check that the first files built are not the ones in the `.o` files already built
  - check that each time a file finishes building, a new `.o` file appears (ie doesnt overwrite an existing one)
  - interrupt the build
- `touch lib/THC/THCStorage.cu`
- start incremental build
  - check that `THCStorage.cu` is one of the first files to be build (note: I'm building on 4-core, so there are 4 'first files' on each build)
  - wait for THCStorage.cu to finish building, THCStorage.o to appear
  - interrupt the build
- `touch lib/THC/THCStorage.h`
  - start incremental build
  - check that `THCStorage.cu` is one of the first files to be build
  - interestingly a bunch of .o files disappear, leading me to think these are for .cu files that depend on THCStorage.h, which seems like a good sign

Nuance:
- when I first created this pull request, I tested on ubuntu 14.04, using gcc-4.8, g++-4.8, cmake-2.8.12.2
- this time, my machine is ubuntu 15.10, gcc-4.9, g++-4.9 (via `CC` and `CXX` vars), and cmake .... 3.2.2
- ironically, or at any rate, interestingly, considering the previous discussion on versioning, it turned out that with my cmake version, a bunch of warnings appeared for policy CMP0054
  - which I fixed by merging line 129 of the cmake-3.2.2 run_nvcc.cmake across to our local copy
  - when I did this, I noticed there are actually a bunch of changes, eg `cusolver` is now added, for CUDA 7.0 and above, in cmake 3.2.2.  This concerns me somewhat, but since it's quite easy to merge things across, and since we are mostly all using ubuntu lts, I think it might kind of work ok, ish.  Since this build option is optional now, I guess we can see how it goes?

hughperkins:
(I suppose another possible approach would be to patch it at directly in cmake instead; that does make a certain amount of sense, and I confess I havent tried that, so I'm not sure how difficult that might be to do)

hughperkins:
Hmmm, looks like a bunch of work... they're not on github... involves signing up for mailing lists and stuff.  There is a maintainer for FindCUDA, 'James Bigler, nvidia.com - jbigler or jamesbigler - gmail dot com' https://cmake.org/Wiki/CMake:Module_Maintainers . Maybe I will email him, see what his thoughts are on the topic.

soumith:
@hughperkins , @borisfom is from NVIDIA, he could probably ping James Bigler even in person :)

borisfom:
I have sent e-mail to James - he is in SLC, so no poking him in person :)
@hughperkins: did not have your e-mail, so copied only Soumith. I'll suggest to James to comment here.

borisfom:
I have merged this PR into my fork of cutorch and tested it. 
My notes:

1. Dependencies work as expected: tried touching .cu and .cuh. Hallelujah!
2. Propagating env variable through luarocks is a real pain - I ended up changing option to be 'ON'.
3. There are other CUDA-enabled modules, notably cunn. I have made changes in my fork of cunn similar to cutorch - but fetching cmake files from cutorch (and relying on opton to be ON) 
4. Given amount of plumbing and some uncertainty if changing cmake module name may have side effects, I would actually vote to rename FindCuda_local back to FindCuda. If we are inclined to have it as an option, we can move it to subdirectory and manipulate CMAKE_MODULE_PATH. 
5. For easier sharing and eliminating duplication, should we put the subdirectory I just proposed, up into extra/ and have FindCUDA being called from extra/ just once ?

hughperkins:
> For easier sharing and eliminating duplication, should we put the subdirectory I just proposed, up into extra/ and have FindCUDA being called from extra/ just once ?

I was actually thinking this.  Especially if James updates FindCUDA to work for this, possibly the easiest way to integrate that into our work could be to bring it down as an 'extra' or 'pkg' repo.

borisfom:
No, why new repo? I would just host it inside extra/ directory in distro. This way (and with moving back to shadowing FindCuda name) no individual repo has to be altered to take advantage of the fix, only distro itself - speaking about maintenance.

hughperkins:
> No, why new repo?

Well, I was thinking, if we just point to a hypothetical git repository hosted by James, we will automatically pull down the latest version of that repo, with latest fixes and so on.  We could also then easily file pull requests and issues against that repo.

borisfom:
I see. Well, it's not that we should anticipate prolonged collaboration on this, really, so I suggest we would just move your fix up to extra/ and keep it shadowing FindCUDA until it's fixed at the source.

hughperkins:
> I see. Well, it's not that we should anticipate prolonged collaboration on this, really, so I suggest we would just move your fix up to extra/ and keep it shadowing FindCUDA until it's fixed at the source.

Yes, alright, sounds reasonable.  Probably should have a separate repo though, rather than distro, to keep distro clean.  I guess this would depend on @soumith's preference(s)?

borisfom:
I think the whole point was to pollute distro with the it :). I don't  think there is much point to have it as separate module - we still need the block adding (conditionally or not) the path to the shadow to CMAKE_MODULE_PATH to be in extra/CMakeLists.txt - might as well have the directory in the same repo (distro) as well.

hughperkins:
Ok, so....
- I'm sort of liaising with James, but he sort of expects me to use [optix](https://developer.nvidia.com/optix) as the reference version of FindCUDA.  It's a 300-1000MB download, behind a registration page :-P  Seems not really ideal for a 2KB line file
- I'm still liaising to figure out what he feels is the recommended way for us to obtain the latest version, install it with cutorch etc.
- As far as the fix, I think he has a slightly more general fix plausibly, which I might look into in detail sometime, but for now, since the current fix seems to work, and I've tested it, leaving as is for now
- more or less in the spirit of installing into pkg/FindCUDA or similar, I've done for now:
  - created a repo FindCUDA, https://github.com/hughperkins/FindCUDA
    - this is an installable rockscpec directory
    - it is sufficient to `luarocks install rocks/findcuda-scm-1.rockspec` to immediately activate incremental builds for both cutorch and cunn :-)
  - I've also hooked it into distro, as a submodule, at https://github.com/hughperkins/distro/tree/add-findcuda , havent created a pullrequest for this yet, we can discuss it in this thread first I think

hughperkins:
Well, I think I will close this pull request for now, and leave the FindCUDA distro around for a bit.  It's easy to install. I'll probably throw a link into the original issues for this issue.

soumith:
sorry to be a bit late to the party, but can we please put FindCUDA forks in cutorch/cunn and not in distro?

cutorch / cunn shouldn't reference anything outside of their directories, especially in distro. Distro is not meant to be a development env in the first place.
Let's just put the modified files into cutorch/cmake and cunn/cmake.

soumith:
i'll add  https://github.com/hughperkins/FindCUDA as a subtree to this repo (cutorch) and to cunn. So no action is necessary from either of you.

hughperkins:
Well... it is actually installable, just like any other torch luarocks module now.  It installs itself into eg ~/torch/install/share/cmake/torch , next to other installed cmake files.  This way, it then works for every cuda torch module, with no modification requried to any of them.  And it's fairly clean.  And it's optional, and doesnt affect people that just want to install torch, not do any development on the actual cutorch and cunn implementation.

soumith:
that's a pretty good idea. thanks.

hughperkins:
Ok :-)

hughperkins:
Update from James Bigler on upstream FindCUDA:

"I probably confused the issue.  All the code you need is in CMake's master and next branches.  In fact, today I just checked in some code that only prints out the missing dependency files when in verbose mode (make VERBOSE=1).  [...]

"After it gets commited into 'master' I'll ask the CMake devs if it will make it into 3.5 RC2."

"My changes won't make it into 3.5, but it will for 3.5.1"

hughperkins:
Update on my own plans: I think I will keep my FindCUDA repo, until such time as James's fixes make into Ubuntu LTS, or that I realize it's easy to install upstream cmake, and automatically get his fixes.  I will probably occasionally mirror James fixes into my repo, as and when some key fix or functionality becomes available.

borisfom:
Sounds like a plan. Thanks for the update!

borisfom:
Also, relying on regular CMake would mean requiring CMAKE version >= 3.5.1 - this may have many side effects and needs to be tested.

hughperkins:
> Also, relying on regular CMake would mean requiring CMAKE version >= 3.5.1 - this may have many side effects and needs to be tested.

Mmmm, right.

borisfom:
Also, my take on the repo topology: I believe it is less intrusive to keep it in the distro repo, under extra/ : them changes are minimal and local to distro repo only: added submodule and a single line change in the install script that installs the FindCuda rock after torch7 but before cutorch.
I think this is exactly what distro repo is for: to provide convenient environment to build all packages.

hughperkins:
Well... I see a couple of arguments against adding it to distro:
1. I see distro as primarily targeted at end-users of torch, rather than people who will hack on the underlying cutorch implementation?  Anyone who has time to figure out how to hack probably has time to figure out how to install an extra package to speed up compilation?
2. It's not entirely certain that the modified FindCUDA will work on every platform.  So making all the end-users install it, who will never modify cutorch ever, seems to be increasing support load, without carrying a clear benefit?

So far, it has zero stars, so there doenst seem to be a huge demand :-P

PR #259, opened 2015-10-08
Add automatic CUDA architecture detection

albanD:
The exact same logic as the cunn version.
More details in the [cunn#151](https://github.com/torch/cunn/pull/151) PR.

### DIFF
--- a/cmake/FindCudaArch.cmake
+++ b/cmake/FindCudaArch.cmake
@@ -1,0 +1,145 @@
+# Borrowed from caffe
+# https://github.com/BVLC/caffe/blob/master/cmake/Cuda.cmake
+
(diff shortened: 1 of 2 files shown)


soumith:
thanks Alban.

soumith:
@albanD can we make it read from an optional environment variable and just fallback to 2.0 ?

albanD:
@soumith yes it would be possible.
Also do you think it would be better when the automatic detection fails (because we failed to compile or run the CUDA code that will give us the architecture) to fall back to only 2.0? Right now in this case we compile for all architectures [here](https://github.com/torch/cutorch/blob/8a1307b403ac79f2c29efb6f078aa97e46b7a18a/cmake/FindCudaArch.cmake#L57).

PR #262, opened 2015-10-12
Fall back to Thrust sort if Torch kernel can't handle the input.

dominikgrewe:
We currently only use Thrust for sorting if we think that it's more efficient than our own method. Also use Thrust for cases that our kernel cannot handle, instead of throwing an error.


soumith:
thanks!

PR #264, opened 2015-10-15
Allow passing no arguments to cutorch.createCudaHostTensor

dominikgrewe:
Still creates a storage because of the non-default allocator.


soumith:
thanks!

PR #285, opened 2015-11-24
Fix tensor index bugs and performance; better testing

wickedfoo:
The old non-contiguous indexSelect kernel was slow, and the code as checked in had a bug (the grid and block were swapped).

Although I cannot explain it by reading the code (it looked correct), we noticed memory corruption behavior from the old contiguous indexSelect kernel as well that only went away when we disabled the contiguous version internally.

I deleted the old kernels and rewrote them with something that is to my mind much simpler, and faster for all except the single outermost contiguous case (though in this case, not slow by much). For a very common case, index() of a 1-d tensor producing a 1-d result, the new kernel is much faster. Once this is accepted, I will replace the other index* kernels using a similar pattern.

The basic idea is that you can handle all dimensions except for the indexed dimension via normal indexing math in a tensor with the indexed dimension being size 1. The indexed dimension's offset (offset += index * stride[index_dim]) is then added in. The indexing is also specialized in the same way that the other pointwise etc. kernels are done as well.

I also added much more strenuous testing: multi-dimensions, large tensors, transposed/tensors with holes. I started to refactor the tensor construction code in test.lua, and will try and sprinkle this around in more tests in subsequent pull requests, since everywhere we should be testing with large + small, contig + non-contig + non-contig-with-holes tensors, especially when there are kernel specializations for various cases.

performance:

image

code for the above perf numbers:
```
local function doTime(f)
   local timer = torch.Timer()
   cutorch.synchronize()
   timer:reset()

   f()
   cutorch.synchronize()
   print(timer:time().real * 1000)
end

local src = torch.CudaTensor(1000, 10, 1000):zero()
local idx = torch.LongTensor({1, 3, 5, 7, 9}):cuda()
doTime(function()
      local out = src:index(2, idx)
end)

local src = torch.CudaTensor(10, 1000, 1000):zero()
local idx = torch.LongTensor({1, 3, 5, 7, 9}):cuda()
doTime(function()
      local out = src:index(1, idx)
end)

local src = torch.CudaTensor(1000, 10, 1000):zero()
local idx = torch.linspace(1, 250, 250):mul(2):cuda()
doTime(function()
      local out = src:index(1, idx)
end)

local src = torch.CudaTensor(2, 5000, 1000):zero()
local idx = torch.LongTensor({2}):cuda()
doTime(function()
      local out = src:index(1, idx)
end)

-- transposed

local src = torch.CudaTensor(1000, 1000, 10):zero():transpose(2, 3)
local idx = torch.LongTensor({1, 3, 5, 7, 9}):cuda()
doTime(function()
      local out = src:index(2, idx)
end)

local src = torch.CudaTensor(10, 1000, 1000):zero():transpose(2, 3)
local idx = torch.LongTensor({1, 3, 5, 7, 9}):cuda()
doTime(function()
      local out = src:index(1, idx)
end)

local src = torch.CudaTensor(1000, 1000, 10):zero():transpose(2, 3)
local idx = torch.linspace(1, 250, 250):mul(2):cuda()
doTime(function()
      local out = src:index(1, idx)
end)

local src = torch.CudaTensor(2, 5000, 1000):zero():transpose(2, 3)
local idx = torch.LongTensor({2}):cuda()
doTime(function()
      local out = src:index(1, idx)
end)


local src = torch.CudaTensor(10, 1000, 1000):zero():transpose(1, 2)
local idx = torch.LongTensor({1, 3, 5, 7, 9}):cuda()
doTime(function()
      local out = src:index(2, idx)
end)

local src = torch.CudaTensor(1000, 10, 1000):zero():transpose(1, 2)
local idx = torch.LongTensor({1, 3, 5, 7, 9}):cuda()
doTime(function()
      local out = src:index(1, idx)
end)

local src = torch.CudaTensor(10, 1000, 1000):zero():transpose(1, 2)
local idx = torch.linspace(1, 250, 250):mul(2):cuda()
doTime(function()
      local out = src:index(1, idx)
end)

local src = torch.CudaTensor(5000, 2, 1000):zero():transpose(1, 2)
local idx = torch.LongTensor({2}):cuda()
doTime(function()
      local out = src:index(1, idx)
end)

-- sparse 1-d

local src = torch.CudaTensor(10000000):zero()
local idx = torch.linspace(1, src:size(1) / 2, src:size(1) / 2):mul(2):cuda()
doTime(function()
      local out = src:index(1, idx)
end)
```

### DIFF
--- a/lib/THC/THCReduceApplyUtils.cuh
+++ b/lib/THC/THCReduceApplyUtils.cuh
@@ -33,21 +33,22 @@ struct CopyOp {
   }
 };
 
-enum NoCollapseMode { NoCollapseDims };
-
 // CUDA kernel argument that defines tensor layout
 template <typename IndexType>
 struct TensorInfo {
   // Extracts size/stride information for the kernel.
-  // Successive dimensions can be collapsed if the size/strides match
-  // up and thus there are no holes between the dimensions. This is used
-  // to reduce the complexity of the problem.
   // The optional `reduceDim` indicates a reduction dimension for the
-  // given tensor, so that the output size for this dimension will be 1.
+  // given tensor, so that the resulting size for this dimension will be 1.
   TensorInfo(THCState* state, THCudaTensor* t, int reduceDim = -1);
 
-  // Doesn't collapse runs of contiguous dimensions.
-  TensorInfo(THCState* state, THCudaTensor* t, NoCollapseMode);
+  // Collapses all runs of successive dimensions if the size/strides
+  // match up within the run and there are no holes between the
+  // dimensions.
+  // If excludeDim is set (not -1), then excludeDim will not be
+  // collapsed with any other dimension.
+  // Function returns the new dimension index that excludeDim maps to,
+  // since the collapsed dimensions are <= the input dimensions.
+  int collapseDims(int excludeDim = -1);
 
   // Contiguous tensors of more than one dimension are collapsed down
   // to one tensor
@@ -64,24 +65,38 @@ struct TensorInfo {
 template <typename IndexType>
 TensorInfo<IndexType>::TensorInfo(THCState* state,
                                   THCudaTensor* t,
-                                  int reduceDim)
-    : data(NULL), dims(0) {
-  int origDims = THCudaTensor_nDimension(state, t);
-  assert(origDims <= MAX_CUTORCH_DIMS);
-  assert(reduceDim < origDims);
-
+                                  int reduceDim) {
   data = THCudaTensor_data(state, t);
+  dims = THCudaTensor_nDimension(state, t);
+  assert(dims <= MAX_CUTORCH_DIMS);
 
-  // Count the number of successive dimensions that can be collapsed, from
-  // innermost to outermost.
-  int numCollapsed = 0;
+  for (int i = 0; i < dims; ++i) {
+    sizes[i] = THCudaTensor_size(state, t, i);
+    strides[i] = THCudaTensor_stride(state, t, i);
+  }
 
+  assert(reduceDim == -1 || (reduceDim < dims && reduceDim >= 0));
+
+  if (reduceDim != -1) {
+    sizes[reduceDim] = 1;
+  }
+}
+
+template <typename IndexType>
+int
+TensorInfo<IndexType>::collapseDims(int excludeDim) {
   // Find the innermost dimension not of size 1, since dimensions of size 1 are
   // collapsible.
   int firstNonOneDim = -1;
 
-  for (int i = origDims - 1; i >= 0; --i) {
-    if (THCudaTensor_size(state, t, i) != 1 && i != reduceDim) {
+  for (int i = dims - 1; i >= 0; --i) {
+    if (i == excludeDim) {
+      // We cannot collapse this dimension, even if it is size 1
+      firstNonOneDim = i;
+      break;
+    }
+
+    if (sizes[i] != 1) {
       firstNonOneDim = i;
       break;
     }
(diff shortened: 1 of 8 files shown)


wickedfoo:
ping

lib/THC/THCReduceApplyUtils.cuh
```diff
@@ -64,13 +73,31 @@ struct TensorInfo {
...
+    strides[i] = THCudaTensor_stride(state, t, i);
+  }
+
+  assert(reduceDim == -1 || (reduceDim < dims && reduceDim >= 0));
+
+  if (reduceDim != -1) {
+    sizes[reduceDim] = 1;
+  }
+
+  collapseDims();
```
dominikgrewe: I wonder if we should let the user collapse dimensions explicitly. Then there's no need for the NoCollapseDims options. What do you think?

lib/THC/THCReduceApplyUtils.cuh
```diff
@@ -91,78 +124,117 @@ TensorInfo<IndexType>::TensorInfo(THCState* state,
...
   for (int i = firstNonOneDim - 1; i >= 0; --i) {
-    long sizeOuter = (i == reduceDim) ? 1 : THCudaTensor_size(state, t, i);
-    long strideOuter = THCudaTensor_stride(state, t, i);
-
-    // The next outermost dimension can be skipped if size 1
-    if (sizeOuter == 1) {
-      ++numCollapsed;
-      continue;
-    }
+    long sizeOuter = sizes[i];
```
dominikgrewe: Can you add a comment saying that this loop only determines how many dimensions can be collapsed, but it doesn't collapse anything? Similarly, a comment on the loop below saying that it collapses the dimensions would be helpful.

lib/THC/THCTensorIndex.cu
```diff
@@ -362,48 +372,90 @@ void THCudaTensor_indexSelect(THCState *state, THCudaTensor *res, THCudaTensor *
...
 
-    int blockX = std::min(THCCeilDiv(nIndex, 4L), 65535L);
-    int blockY = std::min(THCCeilDiv(stride, 128L), 65535L);
+  // The `src` is partitioned into two parts:
+  // -the number of slices over which we are indexing, which is the
+  // total size of the tensor ignoring dimension `dim`;
+  // -the number of indices we are indexing over each of those slices,
+  // which is the total size of the tensor `indices`
+  long totalSize = THCudaTensor_nElement(state, dst);
+  long numSlices = totalSize / numIndices;
```
dominikgrewe: Would `sliceSize` be a better name? When I see `numSlices`, I feel it should be the same as `numIndices`.

dominikgrewe: I guess it's a question of terminology. For me, each index selects a slice from the source tensor.

lib/THC/THCTensorIndex.cu
```diff
@@ -257,78 +257,87 @@ void THCudaTensor_indexFill(THCState *state, THCudaTensor *res_, int dim, THCuda
...
-)
-{
-  for (unsigned long linearIndex = blockIdx.x * blockDim.x + threadIdx.x;
+// We prefer this kernel to balance parallelism across index points
+// and slices otherwise.
+// This kernel in fact works for all choices of problem size, but if
+// the number of indices chosen is small, then the
+// indexSelectSmallIndex kernel is a better choice to reduce memory
+// accesses.
+template <typename IndexType, int DstDim, int SrcDim, int IdxDim, bool IndexInnermost>
```
dominikgrewe: `IndexInnermost` is false at all call sites. Remove it?

dominikgrewe:
Sorry for the delay. Just a few minor comments, otherwise it looks good to me. Thanks!

lib/THC/THCTensorIndex.cu
wickedfoo: Yes, I added this as an experiment. I was thinking about specializing what the index goes over innermost for better performance based on some heuristic, but I'll just remove it.

lib/THC/THCTensorIndex.cu
wickedfoo: works for me, I'll reword it.

lib/THC/THCReduceApplyUtils.cuh
wickedfoo: sure

lib/THC/THCReduceApplyUtils.cuh
wickedfoo: I'll change all the callsites in cutorch that use the default behavior to collapse it explicitly, and remove the separate option.

wickedfoo:
updated with the requested changes; changed all TensorInfo constructor callsites to call collapseDims where applicable

soumith:
Jeff ftw!

PR #286, opened 2015-11-25
Add isSetTo to cutorch

jacobmenick:
Identical behavior to https://github.com/torch/torch7/pull/467.


soumith:
will look at PRs in two days time. Sorry for the delay and non-response. Maybe others will reply sooner.

soumith:
Thanks Jacob.

PR #294, opened 2016-01-07
Implementation of totable for CUDA tensors and storages.

dominikgrewe:


dominikgrewe:
Fixes #161

dominikgrewe:
Added a missing call to `free`. Any other comments on this CL?

soumith:
Thanks!

Issue #268, opened 2015-10-22
Bug in sort when falling back to Thrust


fmassa:
This bug appeared when generalizing sort to handle all input cases, https://github.com/torch/cutorch/pull/262.
Indeed, it seems that `THCudaTensor_sortImplThrust` can't handle some cases properly.

```lua
a = torch.rand(192000,2):float() 
b = a:sort(1,false)
ac = a:cuda()
-- the following did not to work before #262
-- asserting CUTORCH_DIM_WARNING
bc = ac:sort(1,false)
bcc = bc:float()

print(b:eq(bcc):all())
-- prints false

print((b-bcc):abs():max())
-- prints 0.99831330776215 in my test
```

I'm using CUDA 6.5.
Can you reproduce this problem ?
cc @dominikgrewe

fmassa:
Just tested it using CUDA 7.0 (both on Ubuntu 14.04 latest cutorch), and it also fails.
(I'm mentioning CUDA versions because of [those](https://github.com/torch/cutorch/blob/master/lib/THC/THCTensorSort.cu#L414) checks).

szagoruyko:
can reproduce

dominikgrewe:
Yes, looks like the Thrust function cannot handle non-contiguous slices. I suggest for now we just abort in this case (as we did before) and then think of a better solution, unless someone has a better idea.

wickedfoo:
the fallback is still broken along with other broken sort issues, see https://github.com/torch/cutorch/issues/291

I think it can be fixed though. I have the other things except the Thrust fallback fixed locally.

wickedfoo:
this is fixed now

wickedfoo:
please close this issue

Issue #291, opened 2015-12-22
sort() broken in several ways


wickedfoo:
x cannot sort among dimension of size 1 (effectively a no-op except returning an indices list filled with 1s)
x cannot deal with non-contiguous previously allocated indices/sorted results (via torch.sort())
x something seriously broke along the way from initial implementation somehow. Everywhere something like this occurs:
```
    TensorInfo<unsigned int> sortedInfo(state, sorted, dim);
    sortedInfo.collapseDims();
```
needs to be instead:
```
    TensorInfo<unsigned int> sortedInfo(state, sorted, dim);
    sortedInfo.sizes[dim] = 1;
    sortedInfo.collapseDims(dim);
```
(there's a ghost of prior code with a reference to "dropDim" that this is about, but is no longer in the code)

x Thrust fallback sort should check not only that the sorting dimension is contiguous, but that the indices dimension is contiguous. e.g., t = torch.CudaTensor(2, 50000):normal():transpose(1, 2):sort(1) is broken.

One can do segmented sorts in Thrust with some finagling, and I think you can get the iterators to handle strides. I tried a segmented sort before in the THCSort code, but for small sizes it was slower than the bitonic thing.

Thus, I think it is possible to fix the Thrust fallback issue, and actually make cutorch sort work for all inputs, regardless of size, dims or stride.

Fixes for the other issues coming soon.

wickedfoo:
fixed now

PR #296, opened 2016-01-07
top-k implementation + sort works for all cases now

wickedfoo:
This pull request contains:
-an implementation of `values, indices = tensor:topk(k, dim, dir, sorted)` for CudaTensor via efficient radix selection.
-fixing `sort` for all possible inputs, regardless of size, stride, transpositions, holes, whatever.
-better testing for `sort`, and more re-factoring so that we can eventually use test tensors with holes/transpositions/multi-dimensions in all tests.

The top-k elements along slices defined by dimension `dim` are sorted are found and returned using radix selection (radix size of 4 is used here, it seems faster than radix-2 or radix-16 in my tests; the radix 2^k is possible where k is an even divisor of the total number of bits in the value being selected, so k=1, 2, 4, 8, 16, .., but there are 2^k buckets defined, so radix sizes of 2^1, 2^2, 2^4 are the only real feasible ones).

Radix selection is possible for floating-point numbers by bit manipulation, so that if float32 f1 < f2, then convert(f1) < convert(f2) where convert(f) produces an unsigned int and unsigned int comparison is used. Radix selection for top-k requires multiple passes over each input slice, but no scratch space (beyond registers or shared memory) are used in the process.

By default, the elements selected are returned unsorted. If sorting is desired, then a separate sort pass that performs the sorting of key/value pairs inplace is performed.

Because `topk()` uses the sorting mechanism, and I wanted a completely in-place implementation, I refactored the `sort()` code to sort on a key/value tensor pair. Even though there is an additional read and slightly more shared memory used for this sorting implementation, it is slightly faster/about the same/slightly slower than the old sort implementation for many sizes; not enough either way (+/- 3% or so) to be of real concern. The old `values, indices = t:sort(...)` API now uses this one (key, value) sort implementation so I don't need two implementations.

The old sort code was broken in many ways: failed on large numbers of slices of size > 2048, failed on non-contiguous index/sorted value result writes (since the result tensors can be non-contiguous as passed by `torch.sort(resval, resind, ...)` and other failure modes (tensor collapsing was broken, etc.). I now have a backup implementation that uses Thrust to perform a segmented vectorized sort that is the fallback case, and handles all possible sizes/dimensions/strides. In cases where the sorting dimension is not innermost contiguous or where any of the tensors (input, resval or resind) are not contiguous, memory allocations and copies are performed. What is important is that `sort()` now works for all possible inputs and no longer has any problems. The sort is very fast for cases where the slice to sort is <= 2048 elements, since all work is done in shared memory, and is reasonably fast for the Thrust backup when no memory is allocated.

Added new sort and top-k tests that test all manners of sizes and non-contiguous cases. The sort and top-k tests are slow, but there are lots of corner cases in the implementation (slices with size <= 2048 and > 2048 etc.) The cutorch test as a whole needs to get a lot more bulletproof and test many dimension sizes and strides, since with all of the template specialization in the cutorch code, there are many corner cases. Unfortunately to test many of them in a randomized fashion will require test.lua to take longer to run.

The sort and top-k code will work for compute capability 2.0+ (the only interesting intrinsic that I use is `__ballot()` for prefix sums of binary flags). I think CC 2.0+ is our minimum for cutorch anyways (1.x is ancient).

I did add a ldg feature that for 3.5+ code will use the __ldg() intrinsic, but will just use a normal deference for compute capability <3.5.

### DIFF
--- a/lib/THC/THCTensorSort.cu
+++ b/lib/THC/THCTensorSort.cu
@@ -1,4 +1,5 @@
 #include "THCReduceApplyUtils.cuh"
+#include "THCSortUtils.cuh"
 #include "THCTensorCopy.h"
 
 #include <thrust/device_ptr.h>
@@ -6,13 +7,6 @@
 #if CUDA_VERSION >= 7000
 #include <thrust/system/cuda/execution_policy.h>
 #endif
-
-template <typename IndexType, int Power2SortSize>
-__device__ __forceinline__ IndexType
-getSortSliceLinearIndex() {
-  // linear block ID -> slice we are sorting (one per block)
-  return getLinearBlockId<IndexType>();
-}
 
 // Returns 2^(ceil(lg(n)) from Stanford bit twiddling hacks
 unsigned long nextHighestPowerOf2(unsigned long n) {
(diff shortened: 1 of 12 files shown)


wickedfoo:
this fixes

https://github.com/torch/cutorch/issues/268
https://github.com/torch/cutorch/issues/291

lib/THC/THCTensorSort.cu
```diff
@@ -28,303 +22,242 @@ unsigned long nextHighestPowerOf2(unsigned long n) {
...
-      case 2:                                   \
-      HANDLE_CASE(TYPE, A, 2);                  \
-      break;                                    \
-      case 1:                                   \
-      HANDLE_CASE(TYPE, A, 1);                  \
-      break;                                    \
-      default:                                  \
-      assert(false);                            \
-    }                                           \
+#define HANDLE_SORT_CASE(TYPE, A)                       \
```
nicolasvasilache: Consider using boo... err never mind

wickedfoo: yeah i know what you were going to say, but nope :)

wickedfoo: if it were C++11, constexpr and some template magic can do this macro expansion crap, but as a series of successive `if` statements rather than a switch.

wickedfoo: but that code would be more difficult to write and less readable than the macro crap i think

soumith:
@nicolasvasilache does this look good to you to merge?

nicolasvasilache:
@soumith yeah looks good and I like the random unit tests for corner cases.
Good to go from my POV.

wickedfoo:
ping @soumith

soumith:
Sorry for the delay.

PR #300, opened 2016-01-12
Adding cuBLAS matrix inverse

egonina:
torch.inverse requires MAGMA. This adds cuBLAS getrf and getri bindings s.t. we can compute matrix inverse directly with cuBLAS without MAGMA.

moved the inverse test outside of cutorch.magma block


soumith:
cc: @colesbury

soumith:
Thank you Ekaterina!

PR #312, opened 2016-01-28
Reduce tensor size for topk test

GeorgOstrovski:
This particular test (top-k) can be rather slow - I suggest reducing the used tensor sizes to avoid test durations of 10+ minutes.


soumith:
Thanks!

PR #315, opened 2016-02-04
Fixed MAKE_C_FLAGS propagation

borisfom:
Setting MAKE_C_FLAGS should always add, not overwrite, flags.
In this particular case, Android make fails as the --sysroot argument gets lost.


soumith:
Thanks Boris.

PR #317, opened 2016-02-05
Fixing bogus include

borisfom:
TH_INCLUDE_PATH does not exists! The one I changed it too, may not be needed either.


hughperkins:
I think you should remove the FindCUDA commits now ;-)  And just install like `git clone https://github.com/hughperkins/FindCUDA.git; cd FindCUDA; luarocks make rocks/findcuda-scm-1.rockspec`

PR #321, opened 2016-02-10
add setShowAllocations

hughperkins:
This allows showing any CUDA allocations and frees, to ensure that they are not being done inadvertently.

Use like:

```
require 'cutorch'

cutorch.setShowAllocations(1)

print('create tensor')
a = torch.CudaTensor(3,2):uniform()

print('print a')
print('a', a)

print('add 1 to a')
a:add(1)

print('contiguous tensor')
a:contiguous()

print('reshape tensor')
a:reshape(2,3)

print('reshape tensor')
a:reshape(2,3)
```

Output:
```
Activating showAllocations
create tensor
Alloc CUDA storage 24 bytes
print a
a        0.8305  0.9278
 0.4579  0.3724
 0.6263  0.6756
[torch.CudaTensor of size 3x2]

add 1 to a
contiguous tensor
reshape tensor
Alloc CUDA storage 24 bytes
reshape tensor
Alloc CUDA storage 24 bytes
Freeing CUDA storage 24 bytes
Freeing CUDA storage 24 bytes
Freeing CUDA storage 24 bytes
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,6 +19,7 @@ This new tensor type behaves exactly like a `torch.FloatTensor`, but has a coupl
 - `idx = cutorch.getDevice()` : Returns the currently set GPU device index.
 - `count = cutorch.getDeviceCount()` : Gets the number of available GPUs.
 - `freeMemory, totalMemory = cutorch.getMemoryUsage(devID)` : Gets the total and free memory in bytes for the given device ID.
+-  setShowAllocations([true|false])  : if true, prints a message each time CUDA allocation occurs
 - `cutorch.seed([devID])` - Sets and returns a random seed for the current or specified device.
 - `cutorch.seedAll()` - Sets and returns a random seed for all available GPU devices.
 - `cutorch.initialSeed([devID])` - Returns the seed for the current or specified device
--- a/init.c
+++ b/init.c
@@ -691,6 +691,21 @@ static int cutorch_getMemoryUsage(lua_State *L) {
   return 2;
 }
 
+static int cutorch_setShowAllocations(lua_State *L)
+{
+  THCState *state = cutorch_getstate(L);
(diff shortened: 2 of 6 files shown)


apaszke:
Why didn't you just hook to [`THCudaMalloc` or `THCudaFree`](https://github.com/torch/cutorch/blob/master/lib/THC/THCGeneral.c#L605-L621)?
Also, I think it'd be better to pass `true` or `false` rather than `1` or `0` to the Lua function.

hughperkins:
(changed to boolean, still need to address moving hook to THCudaMalloc/THCudaFree)

apaszke:
Also, I think I'd consider printing such diagnostic info to stderr rather than to stdout.

apaszke:
And I'm not even sure if it's really necessary. You can attach a CUDA profiler to torch and see exactly when `cudaMalloc` is used :confused:

hughperkins:
> re: stderr

=> ok, reasonable

> re: hooking into cudamalloc

one issue with this, is that in the THCudaFree, we dont know how many bytes we are freeing?

This makes it hard to track which allocations have actually been freed or not

> You can attach a CUDA profiler to torch and see exactly when cudaMalloc is used

Oh?  I simply ported what I use on the opencl side.  Do you have a url for an easy-to-use cuda profiler, that will work on Ubuntu 14.04 and 15.10?

apaszke:
You can run `nvprof -o out.prof th script.lua` and then open out.prof in nVIDIA Visual Profiler. All this software should be included in your CUDA package already (you can launch Visual Profiler with `nvvp`). It's pretty cool :smile:

hughperkins:
Cool.  I'll try that, and see how that goes. I'll close this for now :-)

Issue #323, opened 2016-02-11
transpose causes Apply1 to be 20 times slower


hughperkins:
transpose causes Apply1 to be 20 times slower.  Even though, except for eg tril, almost all Apply1 ops, with exception eg of `tril` are invariant to transpose.  So, with following code:

```
local time = nil
for it=1,3 do
  print('==== it', it, '==============')

  a = torch.CudaTensor(side, side)
  collectgarbage()
  cutorch.synchronize()
  sys.tic()
  a = a:t()
  -- start, ie assume the tensor arrives transposed
  a:tanh()
  cutorch.synchronize()
  time = sys.toc()
  print('tanh without re-transpose, size', side * side * 4 / 1024 /1024, 'MB =', time, 'seconds')

  a = torch.CudaTensor(side, side)
  collectgarbage()
  cutorch.synchronize()
  sys.tic()
  a = a:t()
  -- start, ie assume the tensor arrives transposed
  a = a:t()
  a:tanh()
  a = a:t()
  cutorch.synchronize()
  time = sys.toc()
  print('tanh with re-transpose, size', side * side * 4 / 1024 /1024, 'MB =', time, 'seconds')


  a = torch.CudaTensor(side, side)
  collectgarbage()
  cutorch.synchronize()
  sys.tic()
  a = a:t()
  -- start, ie assume the tensor arrives transposed
  a:neg()
  cutorch.synchronize()
  time = sys.toc()
  print('neg without re-transpose, size', side * side * 4 / 1024 /1024, 'MB =', time, 'seconds')

  a = torch.CudaTensor(side, side)
  collectgarbage()
  cutorch.synchronize()
  sys.tic()
  a = a:t()
  -- start, ie assume the tensor arrives transposed
  a = a:t()
  a:neg()
  a = a:t()
  cutorch.synchronize()
  time = sys.toc()
  print('neg with re-transpose, size', side * side * 4 / 1024 /1024, 'MB =', time, 'seconds')
end
```

Output:
```
==== it	1	==============
tanh without re-transpose, size	256	MB =	0.67421507835388	seconds
tanh with re-transpose, size	256	MB =	0.036809921264648	seconds
neg without re-transpose, size	256	MB =	0.68109512329102	seconds
neg with re-transpose, size	256	MB =	0.036545038223267	seconds
==== it	2	==============
tanh without re-transpose, size	256	MB =	0.67690086364746	seconds
tanh with re-transpose, size	256	MB =	0.036709070205688	seconds
[...]

PR #324, opened 2016-02-11
autotranspose to stride 1 is final dimension, in apply1

hughperkins:
Addresses https://github.com/torch/cutorch/issues/323

Test case for timings:

```
require 'sys'
require 'cutorch'

local side = 8 * 1024
local time = nil
for it=1,3 do
  print('==== it', it, '==============')

  a = torch.CudaTensor(side, side)
  collectgarbage()
  a = a:t()
  cutorch.synchronize()
  sys.tic()
  -- start, ie assume the tensor arrives transposed
  a:tanh()
  cutorch.synchronize()
  time = sys.toc()
  print('tanh without re-transpose, size', side * side * 4 / 1024 /1024, 'MB =', time, 'seconds')

  a = torch.CudaTensor(side, side)
  collectgarbage()
  a = a:t()
  cutorch.synchronize()
  sys.tic()
  -- start, ie assume the tensor arrives transposed
  a = a:t()
  a:tanh()
  a = a:t()
  cutorch.synchronize()
  time = sys.toc()
  print('tanh with re-transpose, size', side * side * 4 / 1024 /1024, 'MB =', time, 'seconds')

  a = torch.CudaTensor(side, side)
  collectgarbage()
  a = a:t()
  cutorch.synchronize()
  sys.tic()
  -- start, ie assume the tensor arrives transposed
  a:neg()
  cutorch.synchronize()
  time = sys.toc()
  print('neg without re-transpose, size', side * side * 4 / 1024 /1024, 'MB =', time, 'seconds')

  a = torch.CudaTensor(side, side)
  collectgarbage()
  a = a:t()
  cutorch.synchronize()
  sys.tic()
  -- start, ie assume the tensor arrives transposed
  a = a:t()
  a:neg()
  a = a:t()
  cutorch.synchronize()
  time = sys.toc()
  print('neg with re-transpose, size', side * side * 4 / 1024 /1024, 'MB =', time, 'seconds')
end
```

Output, on K520, before PR:
```
tanh without re-transpose, size 256     MB =    0.17799401283264        seconds
tanh with re-transpose, size    256     MB =    0.0059909820556641      seconds
neg without re-transpose, size  256     MB =    0.17719101905823        seconds
neg with re-transpose, size     256     MB =    0.0050539970397949      seconds
```

After PR:
```
tanh without re-transpose, size 256     MB =    0.0051589012145996      seconds
tanh with re-transpose, size    256     MB =    0.0051488876342773      seconds
neg without re-transpose, size  256     MB =    0.0050041675567627      seconds
neg with re-transpose, size     256     MB =    0.0050230026245117      seconds
```

### DIFF
--- a/lib/THC/THCApply.cuh
+++ b/lib/THC/THCApply.cuh
@@ -197,6 +197,14 @@ bool THCudaTensor_pointwiseApply1(THCState* state,
   // additional overhead.
   if (THC_canUse32BitIndexMath(state, a)) {
     TensorInfo<unsigned int> aInfo(state, a);
+
+    // if the final dimension is not the one with stride 1, swap with the one that is,
+    // if such a dimension exists
+    int smallestStrideDim = aInfo.getSmallestStrideDim();
+    if(smallestStrideDim != -1 &&
+        smallestStrideDim != aInfo.dims - 1) {
+      aInfo.transpose(smallestStrideDim, aInfo.dims - 1);
+    }
     aInfo.collapseDims();
 
     HANDLE_A_CASE(unsigned int, aInfo.dims);
--- a/lib/THC/THCReduceApplyUtils.cuh
+++ b/lib/THC/THCReduceApplyUtils.cuh
@@ -16,21 +16,12 @@
 #define CUTORCH_DIM_WARNING "tensor too large or too many (>" \
   CUTORCH_STR(MAX_CUTORCH_DIMS) ") dimensions"
 
+
+// Class definition ==========================================================
+
 // Enum that indicates whether tensor arguments are read/write or
 // read-only
 enum TensorArgType { ReadWrite, ReadOnly };
-
-// Copy operator for the pointwise apply kernel
-template <typename T>
-struct CopyOp {
-  __device__ __forceinline__ void operator()(T* dst, T* src) {
-#if __CUDA_ARCH__ >= 350
-    *dst = __ldg(src);
-#else
-    *dst = *src;
-#endif
-  }
-};
 
 // CUDA kernel argument that defines tensor layout
 template <typename IndexType>
@@ -49,6 +40,10 @@ struct TensorInfo {
   // since the collapsed dimensions are <= the input dimensions.
   int collapseDims(int excludeDim = -1);
 
+  void transpose(int a, int b);
+
+  int getSmallestStrideDim();
+
   // Contiguous tensors of more than one dimension are collapsed down
   // to one tensor
   __host__ __device__ inline bool isContiguous() const {
@@ -61,6 +56,63 @@ struct TensorInfo {
   int dims;
 };
 
+// transversal methods, GPU + CPU, both ok ====================================
+
+// Translate a linear index for the apply to a float* offset;
+// specialized on `Dims` to reduce nvcc compilation time
+template <typename IndexType, int Dims>
+struct IndexToOffset {
+  static __host__ __device__ IndexType get(
+    IndexType linearId,
+    const TensorInfo<IndexType>& info) {
+    IndexType offset = 0;
+
+    // Use static dims
+    for (int i = Dims - 1; i >= 0; --i) {
+      IndexType curDimIndex = linearId % info.sizes[i];
+      IndexType curDimOffset = curDimIndex * info.strides[i];
+      offset += curDimOffset;
+
+      if (i > 0) {
+        linearId /= info.sizes[i];
+      }
+    }
+
+    return offset;
+  }
+};
+
+template <typename IndexType>
+struct IndexToOffset<IndexType, -2> {
+  static __forceinline__ __host__ __device__ IndexType
+    get(IndexType linearId, const TensorInfo<IndexType>& info) {
+    return linearId;
+  }
+};
+
+template <typename IndexType>
+struct IndexToOffset<IndexType, -1> {
+  static __forceinline__ __host__ __device__ IndexType
+    get(IndexType linearId, const TensorInfo<IndexType>& info) {
+    IndexType offset = 0;
+
+    // Use dynamic dims
+    for (int i = info.dims - 1; i >= 0; --i) {
+      IndexType curDimIndex = linearId % info.sizes[i];
+      IndexType curDimOffset = curDimIndex * info.strides[i];
+      offset += curDimOffset;
+
+      linearId /= info.sizes[i];
+    }
+
+    return offset;
+  }
+};
+
+// CPU methods ===============================================================
+// runs on the CPU, usually prior to sending to GPU
+
+// constructor
 template <typename IndexType>
 TensorInfo<IndexType>::TensorInfo(THCState* state,
                                   THCudaTensor* t,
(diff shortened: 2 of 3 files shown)


hughperkins:
Hmmm, I wonder if the transpose should be done prior to the checking for contiguity, and so on?

dominikgrewe:
Shouldn't this work for all point-wise operations? The `Op` being passed just receives a pointer to the value, but not its position. So your change only changes the mapping from CUDA threads to tensor elements.

hughperkins:
> Shouldn't this work for all point-wise operations?

Ah, having just re-read through how `tril` works, it looks like you are right.  Basically, it works out its position in the tensor, by looking at the address of its location within the tensor data, rather than looking at the thread's linearIndex, right?

So, is it fair to say I can remove transposeInvariant?  Are there any other use-cases we should consider?

By the way, it occurred to me that maybe we should just entirely reorder all dimensions, in descending order of stride length, immediately prior to the call to `:collapseDims()`?  Not that I have a good non-n-squared way to do that easily.  But worst case that would be 625 loop iterations anyway (25 squared), and usually more like 25.

dominikgrewe:
> Ah, having just re-read through how tril works, it looks like you are right. Basically, it works out its position in the tensor, by looking at the address of its location within the tensor data, rather than looking at the thread's linearIndex, right?

> So, is it fair to say I can remove transposeInvariant? Are there any other use-cases we should consider?

Yes, I think so.

> By the way, it occurred to me that maybe we should just entirely reorder all dimensions, in descending order of stride length, immediately prior to the call to :collapseDims()? Not that I have a good non-D-squared way to do that easily. But worst case that would be 625 loop iterations anyway (25 squared), and usually more like 25, ie 5 squared.

As long as you use something like insertion sort which is linear for already contiguous tensors, that's not too bad I guess.

dominikgrewe:
Sorry, didn't meant to close this...

hughperkins:
> As long as you use something like insertion sort which is linear for already contiguous tensors, that's not too bad I guess.

Yes.  Not reallly sure the trade-offs here?   A single loop to find the single smallest stride seems fairly cheap, relative to the cost of non-coallesced access.  But a bunch of loops to reduce a bit of GPU calc, and occasionally improve slightly the coallesced access seems less obviously a clear advantage?  Although since it's a loop over a tiny number of values, therefore fully in L1 cache, and can probalby happen whilst the gpu is working on other stuff, maybe it is ok-ish?

dominikgrewe:
I guess the main trade-off is about how much effort you want to invest in speeding up operations that typically don't take up much time overall.

hughperkins:
> I guess the main trade-off is about how much effort you want to invest in speeding up operations that typically don't take up much time overall.

:-D

Well... the development effort seems minimal.  But perhaps the simple single transpose is good for now, to keep things simple.  I dont personally have a use-case for anything other than that for now.

hughperkins:
Update: removed `transposeInvariant` parameter.

wickedfoo:
I think it would be sufficient to just find the dimension with the smallest stride and put that innermost, rather than just looking for stride = 1. A tensor could have small holes (smallest stride = 2 say), where a small number of memory transactions could be made per warp. Reordering everything is probably not worth it, except only in the cases where the smallest stride dimensions are really small in size.

It is worth considering doing this for apply2+ as well but of course that only works if all tensors have the same size/stride layout, so the swap for one tensor works the same on all tensors. For example, adding together two transposed matrices with the same sizes.

I'd also make this a function in TensorInfo itself rather than it being buried in the apply kernel code.

Otherwise, looks good to me.

hughperkins:
The problem is that searching for smallest stride is not `O(n)`, but probably `O(n log(n))`, or maybe `O(N^2)` , if I do a naive implementation, which I probalby would :-P

> It is worth considering doing this for apply2+ as well but of course that only works if all tensors have the same size/stride layout, so the swap for one tensor works the same on all tensors. For example, adding together two transposed matrices with the same sizes.

Yes. I think maybe make a second PR for that perhaps?

> I'd also make this a function in TensorInfo itself rather than it being buried in the apply kernel code.

Ah, that's probably a good idea.  I'll look into that.

hughperkins:
(updated to find smallest stride, rather than stride 1.  To do: factorize into TensorInfo)

hughperkins:
(updated to consider only strides greater than 0)

hughperkins:
Well... for factorizing into TensorInfo... TensorInfo can only 'see' a single tensor.  So, if we want to use in apply2 and apply3, the TensorInfo function wont have the visibility it requires?  It is too low-level, needs to be at a more transverse level, like say ...Apply?

However, the transpose should certainly be in TensorInfo, and probably is.  I'll either add it, or use it.

hughperkins:
- created `void transpose(int a, int b)` method in TensorInfo
- created `int getSmallestStrideDim()` method in TensorInfo
- I also reorged THCReduceApplyUtils.cuh a bit, to group gpu bits together, and cpu bits together.  I understand this might mess with `blame` a bit, so I can undo the reorg if better?

hughperkins:
(Hmmm,random thought, for stride=0 dimensions, what happens to them?  such dimensions should be `squeeze`d out I think?  Does collapse handle this already?
)

hughperkins:
Oh... stride 0 dimensions will cause a reallocation, prior to the `collapse()`, right?

hughperkins:
(Random thought: rather than duplicating a bunch of functionality into TensorInfo, should we perhaps be automaticlaly creating new tensors, from the same underlying storage, at the start of each Apply, then doing collapses and so on on these tensors, and then making TensorInfo be a relatively dumb struct?)

wickedfoo:
I think it's sufficient to just look for the smallest positive stride. The 0 stride tensors are kind of degenerate anyways (especially for write arguments; I think writing into a tensor with any stride of 0 should be an error throughout Torch7 and cutorch, since in a parallel world it is unclear what value you might want to use). Stride 0 is used to repeat and broadcast elements; doing so via stride-0 memory lookups on the GPU is kind of slow in any case. As far as negative strides are concerned, I would cry but I wouldn't be surprised (for the worst) if Torch allows that.

re: operations in the THCudaTensor API itself, I kind of feel that any operation added to that to allow for this kind of collapsing should also be added to all the Torch7 CPU tensors. The manual memory management of Torch tensors (and general abuse of C macros in Torch7) feels rather clunky to me; if Torch7 doesn't need to perform this magic, it makes sense to me to just keep it small in cutorch (and we can iterate on it much faster). Perhaps have functions in TensorInfo that return new TensorInfos by value, so all TensorInfos are immutable.

I'm fine with free-floating functions instead of member functions in TensorInfo too; I was thinking more about keeping the size/stride/collapsing/TensorInfo manipulating code all together, rather than having some of it be in the apply functions.

You can write functions that take a variable number of TensorInfo arguments and collapse them mutually; these make sense as free-floating functions. In C++11 I'd do something like:
```
template <typename... TensorInfos>
std::tuple<TensorInfos...> collapse(TensorInfos args...) { ... }
```

so you don't have to specialize the 1/2/3 etc. case separately. Or in C++98, something like:
```
template <typename IndexType>
void collapse2(TensorInfo<IndexType>& t1, TensorInfo<IndexType>& t2) { ... }
```

or if TensorInfo becomes immutable:
```
void collapse2(const TensorInfo<IndexType>& t1, const TensorInfo<IndexType>& t2,
                        TensorInfo<IndexType>* o1, TensorInfo<IndexType>* o2) { ... }
```

or worst-case, something C-style with variadic arguments but that would require passing TensorInfo<> by pointer.

These functions could look to see if all tensors can be collapsed and transformed together the same way (to move stride-1 dimensions to innermost etc.), perform the work, and return the results.

PR #327, opened 2016-02-13
For discussion: alternative approaches to Apply macros

hughperkins:
For discussion mostly: trying to simplify Apply, so the macros take up less space.  I dont think I clearly achieve this goal, but it is at least different.  Throwing it out for discussion.

Key ideas:
- create a single macro  `EXPAND`, to take any variable, and write it as an appropriate `switch`, substituting in `left` and `right` clauses for each value, and the hard-coded numerical value
- check for contiguity prior to calling this macro, to keep the macro simple, ie we've already replaced eg `a` by `-2`, prior to calling macro.  Not clearly better.  Different at least
- encapsulate each level of variable expansion in a function.  This doesnt really save any space, and its  a bit wordy in its own rigth, so again, not better, just ... different

### DIFF
--- a/lib/THC/THCApply.cuh
+++ b/lib/THC/THCApply.cuh
@@ -14,6 +14,29 @@
 // Threads per block for our apply kernel
 #define THC_APPLY_THREADS_PER_BLOCK 32 * 16
 
+#ifndef COMMA
+#define COMMA ,
+#endif
+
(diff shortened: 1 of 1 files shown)


soumith:
@wickedfoo is on vacation till end of next week. let's pick this up after he's back.

Issue #329, opened 2016-02-15
Failure detected in __THCudaCheck might not survive stack unwind


hughperkins:
I got a seg fault something like the following.  The cause of this seg fault is outside the scope of htis issue:
```
luajit: cuda runtime error (77) : an illegal memory access was encountered at /home/ubuntu/git/cutorch/lib/THC/generic/THCStorage.c:170
```
(Note that line number might not exactly match master THCStorage.c, because I added a ton of `printf`s whilst trying to debug the issue.)

After instrumenting THCStorage.c a bit, it looked like it was crashing during a Storage_free:
```
Storage_free x self=171492384
...cudafree 30136074240
```
After this line, the seg fault occurred, so it seemed maybe some issue with the free.

In fact, the error happened about 15,000 lines of debug earlier, inside THCTensorMath.cu:
```
thcudacheck FAIL file=/home/ubuntu/git/cutorch/lib/THC/THCTensorMath.cu line=303
```
Unfortunately, this error message was swallowed, because the program segfaulted before printing it.  It appears above after adding some instrumentation to __THCudaCheck:
```
void __THCudaCheck(cudaError_t err, const char *file, const int line)
{
  if(err != cudaSuccess)
  {
    printf("thcudacheck FAIL file=%s line=%i\n", file, line );
    _THError(file, line, "cuda runtime error (%d) : %s", err,
             cudaGetErrorString(err));
  }
}
```
After adding this code, it became obvious that the error was not during the free, but much earlier.  Actually, it was about 15,000 lines of debug statements earlier, but swallowed by the segfault during stack unwind.

I think it could be worth ensuring that the error is printed during __THCudaCheck, in case it doesnt survive subsequent unwind.

soumith:
@hughperkins when you are debugging CUDA errors, the first thing one should do is to switch to synchronous calls, so that you can identify the point of failure.

CUDA_LAUNCH_BLOCKING=1 th [yourscript.lua]

This triggers a THCudaCheck exactly at the point of error.
If the launches are asynchronous, THCudaCheck is only triggered when the launched kernel finally computes, which might be much later.

hughperkins:
Good info. Thanks!

PR #330, opened 2016-02-15
Add printf to __THCudaCheck fail clause, in case stack unwind itself …

hughperkins:
…crashes

Address https://github.com/torch/cutorch/issues/329

### DIFF
--- a/lib/THC/THCGeneral.c
+++ b/lib/THC/THCGeneral.c
@@ -520,6 +520,11 @@ void __THCudaCheck(cudaError_t err, const char *file, const int line)
 {
   if(err != cudaSuccess)
   {
+    static int alreadyFailed = 0;
+    if(!alreadyFailed) {
+      fprintf(stderr, "THCudaCheck FAIL file=%s line=%i error=%i : %s\n", file, line, err, cudaGetErrorString(err));
+      alreadyFailed = 1;
+    }
     _THError(file, line, "cuda runtime error (%d) : %s", err,
              cudaGetErrorString(err));
   }


hughperkins:
Hmmm, although note that this will print a ton of FAILs during unwind.  Maybe I should add a `static` var, and only print hte first failure perhaps?

hughperkins:
Modified to only print the first failure, and then assume we're already unwinding, since THError will already have been called at least once.

soumith:
are you sure this actually helps in improving errors? (I dont think so). See my comment on CUDA_LAUNCH_BLOCKING on the other thread.

hughperkins:
It was a prerequisite for diagnosing https://github.com/torch/cutorch/pull/332 (which took me almost my entire day by the way... of which 3-4 hours was working my way backwards far enough to make the change in this PR; I spent several hours instrumenting Storage, trying to work out why my storage was broken; checking pairs of allocaitons and free.  All a complete waste of time of course, since the root cause was prior to the 8,000 frees...)

soumith:
that sounds horrible, i feel sorry for you :)
If you can give me a stack trace output of with this, and without this -- that'll help me understand how this printf is helpful.

hughperkins:
Original stack trace was simply something like:

```
segmentation error
```
That was it :-P  That was the entire output.

If you opened gdb on this line, with a breakpoint, it just unhelpfully showed a call from lua into storage_free.

After instrumenting Storage a bit, I got the following output:

https://gist.github.com/hughperkins/ce225700d0df80b908a7

After instrumenting CudaCheck, I got:

https://gist.github.com/hughperkins/2332cc8f8227454b3960

... which makes the location of hte problem significantly more obvious.  It's nothing to do with Storage_free in fact ....

hughperkins:
(note: the seg faults occurs right at the end of the logfile, in both cases.  the seg fault terminates the program, and closes the logfile; therefore it's reasonable to suppose that the lsat copule of lines of the log are the cause of the segfault.  Which is kind of true. But misleading)

soumith:
ok thanks for the tracing. I expected the line below, i.e.
_THError(file, line, "cuda runtime error (%d) : %s", err,
               cudaGetErrorString(err));

should do the same thing as your printf, but clearly it does not print file/line number on segfault, weird.

hughperkins:
Yeah, it only prints stuff *after* unwind.  And if unwind crashes... :-(

hughperkins:
(Although plausibly one could modify _THError to do a printf before commencing unwind I suppose?)

soumith:
ok this seems reasonable to merge. but I'd want the cudaGetErrorString to also be present in the printf, can you modify that.

hughperkins:
> ok this seems reasonable to merge. but I'd want the cudaGetErrorString to also be present in the printf, can you modify that.

Ok, sounds good.  I'll update that.

By the way for _THError.... curiously it seems that it *should* do a printf in fact.  Here is _THError, in torch7's general.c:

```
void _THError(const char *file, const int line, const char *fmt, ...)
{
  char msg[2048];
  va_list args;

  /* vasprintf not standard */
  /* vsnprintf: how to handle if does not exists? */
  va_start(args, fmt);
  int n = vsnprintf(msg, 2048, fmt, args);
  va_end(args);

  if(n < 2048) {
    snprintf(msg + n, 2048 - n, " at %s:%d", file, line);
  }

  (*torchErrorHandlerFunction)(msg, torchErrorHandlerData);
}
```
What is torchErrorHandlerFunction?  It's:
```
static __thread void (*torchErrorHandlerFunction)(const char *msg, void *data) = defaultTorchErrorHandlerFunction;
```
And defaultTorchErrorHandlerFunction is:
```
static void defaultTorchErrorHandlerFunction(const char *msg, void *data)
{
  printf("$ Error: %s\n", msg);
  exit(-1);
}
```
Seems like it should printf in fact???

hughperkins:
Is it possible that cudaGetErrorString fails???

soumith:
that's a scary thought, but maybe :)

hughperkins:
It's annoyingly hard to test unfortunately.   I tried creating a test case like:
```
require 'cutorch'

a = torch.CudaTensor(128, 128):fill(-1)
b = torch.CudaTensor(128, 128):fill(-1)
c = torch.CudaTensor(128, 128)
local d = {}
for i=1,10000 do
  d[i] = c:narrow(2,1,50)
end
c:scatter(1, b, c)
c:max()
```
... but doesnt reproduce the seg fault.  The only way I have to reproduce the issue is to run my training program, which is quite slow.

Anywya, I guess I should dig a bit, figure out why THError isnt doing printf and/or if cudaGetErrorString is failing.

hughperkins:
Well, one observation is that one of the first htings that happens when I run the program is that `THSetErrorHandler` is called, so the error handler may not be the one we think it is?

hughperkins:
Ah-hah!  In torch7, utils.c:

```
static void luaTorchErrorHandlerFunction(const char *msg, void *data)
{
  lua_State *L = data;
  luaL_error(L, msg);
}

static int torch_updateerrorhandlers(lua_State *L)
{
  THSetErrorHandler(luaTorchErrorHandlerFunction, L);
  THSetArgErrorHandler(luaTorchArgErrorHandlerFunction, L);
  return 0;
}
```

And `luaTorchErrorHandlerFunction` is a printf-free zone.

hughperkins:
Sooo... we could modify `luaTorchErrorHandlerFunction` in `utils.c` to have a `printf` first, but I suspect that things in a non-cuda world are maybe more well-behaved?  So maybe we just fix it in `THCudaCheck` for now, since maybe the issue doesnt occur in torch7 world?

hughperkins:
Ok:
- added `cudaGetErrorString`
- also made it print to `stderr`

Example output:
```
THCudaCheck FAIL file=/home/ubuntu/git/cutorch/lib/THC/THCTensorMath.cu line=303 error=77 : an illegal memory access was encountered
luajit: cuda runtime error (77) : an illegal memory access was encountered at /home/ubuntu/git/cutorch/lib/THC/generic/THCStorage.c:147
```

soumith:
Thank you so much!

Issue #331, opened 2016-02-15
Max gives invalid indexes when fed infinity


hughperkins:
Test program:
```
require 'cutorch'

a = torch.CudaTensor(10, 10):fill(-1/0)
m, i = a:max(1)
print('i[1][1]', i[1][1])
```
Output:
```
i[1][1] -1
```
This is an invalid index.  This causes a crash in nn.Max module, during backprop, when fed inf in forward prop https://github.com/torch/cunn/issues/210

If all the values are all equal to infinity, there is no unique max, but there again, if all the values were some non-infinite value `i`, what would be max?  Try it:
```
torch.CudaTensor(3):fill(5):max(1)
 5
[torch.CudaTensor of size 1]

 2
[torch.CudaTensor of size 1]
```
... it just returns an arbitrary, but valid, index.  I think that passing all infinities should also return a valid, albeit plausibly arbitrary, index, eg `1`.

PR #332, opened 2016-02-15
change default index for min/max to be a valid index

hughperkins:
Address https://github.com/torch/cutorch/issues/331 , which is root cause of https://github.com/torch/cunn/issues/210


hughperkins:
Note: verified behavior in base `torch7`, and it matches that in this PR.

Test on base torch7:
```
require 'torch'

a = torch.Tensor(10, 10):fill(-1/0)
m, i = a:max(1)
print('i[1][1]', i[1][1])

require 'nn'

net = nn.Max(1)

for it=1,3 do
  out = net:forward(a)
  net:backward(a, torch.Tensor(10):uniform())
end
```
Output:
```
i[1][1] 1
```

soumith:
Thanks Hugh!

Issue #320, opened 2016-02-09
indexCopy and indexAdd are broken


soumith:
X = torch.rand(4, 500):cuda()
C = torch.CudaTensor(2, 500):zero()
indices = torch.LongTensor({1,2,2,2})
C:indexCopy(1, indices, X);
C[1]:eq(X[1]):sum()

256

C[1]:eq(X[1]):sum() should be 500.
This problem also affects indexAdd too (since they use the same indexing logic) ...

Looks like nothing is getting copied beyond the first 256 elements.
cc: @dominikgrewe

soumith:
possibly indexFill as well.

Allan Jabri is prepping a fix.

soumith:
Fix might be in the lines of https://github.com/torch/cutorch/pull/285

wickedfoo:
this should be fixed now

PR #336, opened 2016-02-19
Fix indexing bug in indexCopy, indexAdd, indexFill in Cutorch + tests

ajabri:
Fix indexing bug in indexCopy, indexAdd, indexFill in Cutorch + tests. The indexing logic was broken (and only touched indices <= 256).


soumith:
Fixes https://github.com/torch/cutorch/issues/320

soumith:
Reviewed by @wickedfoo

PR #344, opened 2016-02-27
Add math functions trunc, frac, rsqrt, lerp

andreaskoepf:
- [x] cuda implementation of trunc, frac,rsqrt, lerp
- [x] unit tests


andreaskoepf:
rebased & squashed

Issue #60, opened 2014-10-28
:csub does not exist


soumith:
subtraction

Kaixhin:
Has this somehow been removed? I just got `attempt to call method 'csub' (a nil value)` on a CudaTensor in my program, and can't see the method in the interpreter when I create a CudaTensor. Other tensor types are fine.

soumith:
hmm, weird you are right, csub seems to have been removed somewhere in between. taking a look

soumith:
actually, turns out csub wasn't in that PR, I might've wrongly closed it.

daranday:
I'm experiencing the same issue. Pulled the latest version. February 15, 2016

soumith:
fixed via #351

PR #351, opened 2016-03-03
Fix csub #60

JoostvDoorn:
Fix for issue #60.

### DIFF
--- a/TensorMath.lua
+++ b/TensorMath.lua
@@ -388,6 +388,18 @@ wrap("add",
         {name=real, default=1},
         {name=Tensor}})
 
+
+wrap("csub",
+     cname("sub"),
+     {{name=Tensor, default=true, returned=true, method={default='nil'}},
+      {name=Tensor, method={default=1}},
+      {name=real}},
+     cname("csub"),
+     {{name=Tensor, default=true, returned=true, method={default='nil'}},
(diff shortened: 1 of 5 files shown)


soumith:
Thanks!

TensorMath.lua
```diff
@@ -388,6 +388,18 @@ wrap("add",
         {name=real, default=1},
         {name=Tensor}})
 
+
+wrap("csub",
+     cname("sub"),
```
fmassa: This conflicts with the [sub](https://github.com/torch/torch7/blob/master/doc/tensor.md#tensor-subdim1s-dim1e---dim4s--dim4e) function, right ?

soumith: I think it's okay. the sub is the cname, not the lua name (which is csub). We have it the same in torch7 https://github.com/torch/torch7/blob/master/TensorMath.lua#L259-L268

fmassa: Actually no. `sub` is defined at the [lua level](https://github.com/torch/torch7/blob/master/generic/Tensor.c#L289), so it shouldn't be a problem. But maybe it can be ambiguous

fmassa: Yeah, it's ok. My bad !

JoostvDoorn:
Great I did not break anything, thanks for checking.

Issue #349, opened 2016-03-01
cutorch + Titan Black shouldn't build with sm_52


nicholas-leonard:
**Please do not close this issue until it is resolved. **

I am unable to install cutorch on an ubuntu 14.04 server running cuda 6.5. I get the following error : 

```bash
luarocks install cutorch

Installing https://raw.githubusercontent.com/torch/rocks/master/cutorch-scm-1.rockspec...
Using https://raw.githubusercontent.com/torch/rocks/master/cutorch-scm-1.rockspec... switching to 'build' mode
Cloning into 'cutorch'...
remote: Counting objects: 105, done.
remote: Compressing objects: 100% (100/100), done.
remote: Total 105 (delta 8), reused 43 (delta 3), pack-reused 0
Receiving objects: 100% (105/105), 150.16 KiB | 0 bytes/s, done.
Resolving deltas: 100% (8/8), done.
Checking connectivity... done.
cmake -E make_directory build && cd build && cmake .. -DCMAKE_BUILD_TYPE=Release -DCMAKE_PREFIX_PATH="/home/nicholas14/torch/install/bin/.." -D
CMAKE_INSTALL_PREFIX="/home/nicholas14/torch/install/lib/luarocks/rocks/cutorch/scm-1" && make -j$(getconf _NPROCESSORS_ONLN) install

-- The C compiler identification is GNU 4.8.4
-- The CXX compiler identification is GNU 4.8.4
-- Check for working C compiler: /usr/bin/cc
-- Check for working C compiler: /usr/bin/cc -- works
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Check for working CXX compiler: /usr/bin/c++
-- Check for working CXX compiler: /usr/bin/c++ -- works
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Found Torch7 in /home/nicholas14/torch/install
-- Found CUDA: /usr/local/cuda (found suitable version "6.5", minimum required is "5.5")
-- MAGMA not found. Compiling without MAGMA support
-- Automatic GPU detection failed. Building for all known architectures.
-- Compiling for CUDA architecture: 2.0 2.1(2.0) 3.0 3.5 5.0 5.2
-- Configuring done
-- Generating done
-- Build files have been written to: /tmp/luarocks_cutorch-scm-1-5798/cutorch/build
[...]

PR #353, opened 2016-03-05
fixing gpu arch detection fallback

soumith:
fixes https://github.com/torch/cutorch/issues/349


soumith:
@nicholas-leonard see if this fixes things for you now for cutorch. if so, will push the same fix to cunn.

PR #355, opened 2016-03-07
Add FP16 support (CudaHalfStorage, CudaHalfTensor)

apaszke:
This commit adds basic support for half-precision floating-point torch types. For now these only implement basic operations (the same as e.g. CudaDouble{Storage,Tensor}), but it's enough to allow cuDNN or other libraries to work on half precision weights.

There's still much space for improvement:
* `THCudaHalfTensor_copyAsync*` functions are not implemented
* `THCudaHalfStorage_get` is not implemented (requires a fp16->fp32 conversion function for host - to be added later)
* Any copy/conversion (except ones from Cuda{Tensor, Storage} and CudaHalf{Tensor, Storage}) will allocate a temporary float buffer (on the GPU) that has the same number of elements as source and destination. It's just the simplest way to do that, but it's worth reimplementing it in the future, so memory complexity can be reduced.

### DIFF
--- a/lib/THC/THCHalf.cu
+++ b/lib/THC/THCHalf.cu
@@ -1,0 +1,104 @@
+#include "THCHalf.h"
+#include <thrust/transform.h>
+#include <thrust/execution_policy.h>
+
+struct __half2floatOp {
+  __device__ float operator()(half v) { return __half2float(v); }
+};
+
+struct __float2halfOp {
+  __device__ half operator()(float v) { return __float2half(v); }
+};
+
+void THCFloat2Half(THCState *state, half *out, float *in, long len) {
+  thrust::transform(
+#if CUDA_VERSION >= 7000
+    thrust::cuda::par.on(THCState_getCurrentStream(state)),
+#else
+    thrust::device,
+#endif
+    in, in + len, out, __float2halfOp());
+}
+
+void THCHalf2Float(THCState *state, float *out, half *in, long len) {
+  thrust::transform(
+#if CUDA_VERSION >= 7000
+    thrust::cuda::par.on(THCState_getCurrentStream(state)),
+#else
+    thrust::device,
+#endif
+    in, in + len, out, __half2floatOp());
+}
+
+float THC_half2float(half a)
+{
+  THError("half2float not implemented yet");
+  return 0;
+}
+
+/*
+  Copyright (c) 2015, Norbert Juffa
+  All rights reserved.
+
+  Redistribution and use in source and binary forms, with or without
+  modification, are permitted provided that the following conditions
+  are met:
+
+  1. Redistributions of source code must retain the above copyright
+     notice, this list of conditions and the following disclaimer.
+
+  2. Redistributions in binary form must reproduce the above copyright
+     notice, this list of conditions and the following disclaimer in the
+     documentation and/or other materials provided with the distribution.
+
+  THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
+  "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
+  LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR
+  A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT
+  HOLDER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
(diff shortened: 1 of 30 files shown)


szagoruyko:
:+1:

lib/THC/THCHalf.cu
```diff
@@ -0,0 +1,105 @@
...
+    long i = threadIdx.x + blockIdx.x * blockDim.x;
+    for (; i < len; i += gridDim.x * blockDim.x)
+        out[i] = __half2float(in[i]);
+}
+
+void THCFloat2Half(THCState *state, half *out, float *in, long len) {
+    long num_blocks = (len + NUM_THREADS - 1) / NUM_THREADS;
+    if (num_blocks > MAX_BLOCKS)
+        num_blocks = MAX_BLOCKS;
+    _floatToHalf_kernel<<<num_blocks, NUM_THREADS, 0, THCState_getCurrentStream(state)>>>(out, in, len);
```
soumith: use a thrust:: primitive here, rather than this kernel. they work out much better.
Look at thrust::transform http://docs.nvidia.com/cuda/thrust/#transformations

lib/THC/generic/THCStorage.c
```diff
@@ -17,18 +17,23 @@ int THCStorage_(elementSize)(THCState *state)
...
-real THCStorage_(get)(THCState *state, const THCStorage *self, long index)
+hostreal THCStorage_(get)(THCState *state, const THCStorage *self, long index)
 {
   real value;
   THArgCheck((index >= 0) && (index < self->size), 2, "index out of bounds");
+#ifndef THC_REAL_IS_HALF
   THCudaCheck(cudaMemcpy(&value, self->data + index, sizeof(real), cudaMemcpyDeviceToHost));
-  return value;
+#else
+  THError("Get is not supported for CudaHalfStorage. Please convert to float first.");
```
soumith: can you work a story for this, get is essential and used everywhere. converting to float first is not an option in terms of usability.

soumith:
i made some comments inline, the PR looks good for the most part.

adamlerer:
+1

lib/THC/generic/THCStorage.c
apaszke: I'll try to implement this fp16->fp32 host conversion soon. I can't think of any other way of doing that quickly right now (launching a kernel seems too expensive for such a simple operation).

lib/THC/THCHalf.cu
apaszke: I tried using thrust::transform, but it requires functor's `operator()` to be `__host__ __device__`, so I can't call `__device__ __half2float` from there :confused:

apaszke: Ok, `thrust::transform(thrust::device` did the job

apaszke:
@soumith I modified everything according to your comments.

soumith:
can you remind me why copyAsync is not implemented?

soumith:
also, will leave this here for a couple of days, for folks like @dominikgrewe and @wickedfoo to take a look.

apaszke:
@soumith Sure.
Well, I have been a bit imprecise. `THCudaHalfTensor` implements no functions that take host tensors as arguments, including `copyAsync*` (`THTensor` macro expands to an inexistent `THHalfTensor`). As we discussed earlier it would make sense to add that CPU counterpart, and then it wold be simply a matter of removing `#ifndef`s that exclude these functions.

apaszke:
I modified a small bit in `THCHalf.cu` so that fp16 conversion is performed on a currently selected stream.

```diff
diff --git a/lib/THC/THCHalf.cu b/lib/THC/THCHalf.cu
index 3249a46..e368d2e 100644
--- a/lib/THC/THCHalf.cu
+++ b/lib/THC/THCHalf.cu
@@ -11,11 +11,23 @@ struct __float2halfOp {
 };

 void THCFloat2Half(THCState *state, half *out, float *in, long len) {
-  thrust::transform(thrust::device, in, in + len, out, __float2halfOp());
+  thrust::transform(
+#if CUDA_VERSION >= 7000
+    thrust::cuda::par.on(THCState_getCurrentStream(state)),
+#else
+    thrust::device,
+#endif
+    in, in + len, out, __float2halfOp());
 }

 void THCHalf2Float(THCState *state, float *out, half *in, long len) {
-  thrust::transform(thrust::device, in, in + len, out, __half2floatOp());
+  thrust::transform(
+#if CUDA_VERSION >= 7000
+    thrust::cuda::par.on(THCState_getCurrentStream(state)),
+#else
+    thrust::device,
+#endif
+    in, in + len, out, __half2floatOp());
 }

 float THC_half2float(half a)
```

soumith:
Awesome! Thanks.

PR #359, opened 2016-03-08
Improve inverse test

GeorgOstrovski:
My previous fix of this test, adjusting the numerical precision, wasn't enough to make it stable (it was still failing ~ 1% of the runs). A better solution is to construct the matrix to be well-conditioned, so that inversion is numerically more stable - done here as perturbation of identity matrix.

### DIFF
--- a/test/test.lua
+++ b/test/test.lua
@@ -1677,10 +1677,10 @@ function test.ger()
 end
 
 function test.inverse()
-   local a = torch.randn(5, 5)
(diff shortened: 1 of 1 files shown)


soumith:
Thanks!

sefira:
The dimension of Tensor for test is just 5, too small to find problem.
Actually, I and @jonathanasdf meet a bug when torch.inverse(a_cuda_tensor), [issues 507](https://github.com/torch/cutorch/issues/507)

PR #360, opened 2016-03-09
Adding cross to cutorch

ebetica:
Already in torch, implements same functionality in CUDA.

### DIFF
--- a/lib/THC/THCTensorMath2.cu
+++ b/lib/THC/THCTensorMath2.cu
@@ -660,3 +660,49 @@ void THCudaTensor_randn(THCState *state, THCudaTensor *r_, THLongStorage *size)
   THCudaTensor_resize(state, r_, size, NULL);
   THCudaTensor_normal(state, r_, 0, 1);
 }
+
+struct TensorCrossOp {
+  TensorCrossOp(long sx, long sy, long so) : sx(sx), sy(sy), so(so) {}
+
+  __device__ __forceinline__ void operator()(float* out, float* x, float*y) {
+    out[0 * so] = x[1 * sx] * y[2 * sy] - x[2 * sx] * y[1 * sy];
+    out[1 * so] = x[2 * sx] * y[0 * sy] - x[0 * sx] * y[2 * sy];
+    out[2 * so] = x[0 * sx] * y[1 * sy] - x[1 * sx] * y[0 * sy];
+  }
+
+  const long sx, sy, so;
+};
+
+THC_API void THCudaTensor_cross(THCState *state, THCudaTensor *self, THCudaTensor *x, THCudaTensor *y, int dimension)
+{
+  THAssert(THCudaTensor_checkGPU(state, 3, self, x, y));
+
+  int i;
+  long nd = THCudaTensor_nDimension(state, x);
+  long nelem = THCudaTensor_nElement(state, x);
+  THArgCheck(nd == THCudaTensor_nDimension(state, y), 1, "tensors must have same number of dimensions");
+  for (i = 0; i < nd; i++) {
+    THArgCheck(THCudaTensor_size(state, x, i) == THCudaTensor_size(state, y, i), 1, "dimension %i of x and y does not match", i);
+    if (dimension < 0 && THCudaTensor_size(state, x, i) == 3) {
+      dimension = i; 
+    }
+  }
+
+  THArgCheck(dimension >= 0 && dimension < nd, 3, "dimension %d out of range", dimension+1);
+  THArgCheck(THCudaTensor_size(state, x, dimension) == 3, 3,
+      "dimension %d does not have size 3", dimension+1);
+  THCudaTensor_resizeAs(state, self, x);
+
+  long sx = THCudaTensor_stride(state, x, dimension);
+  long sy = THCudaTensor_stride(state, y, dimension);
+  long so = THCudaTensor_stride(state, self, dimension);
+  THCudaTensor *nx = THCudaTensor_newNarrow(state, x, dimension, 0, 1);
+  THCudaTensor *ny = THCudaTensor_newNarrow(state, y, dimension, 0, 1);
+  THCudaTensor *nself = THCudaTensor_newNarrow(state, self, dimension, 0, 1);
+  if (!THCudaTensor_pointwiseApply3(state, nself, nx, ny, TensorCrossOp(sx, sy, so))) {
+    THArgCheck(false, 2, CUTORCH_DIM_WARNING);
+  }
+  THCudaTensor_free(state, nx);
+  THCudaTensor_free(state, ny);
+  THCudaTensor_free(state, nself);
+}
--- a/TensorMath.lua
+++ b/TensorMath.lua
@@ -693,6 +693,13 @@ for _,name in ipairs({"cmin", "cmax"}) do
          {name=real}})
 end
 
+wrap("cross",
+cname("cross"),
+    {{name=Tensor, default=true, returned=true},
+     {name=Tensor},
+     {name=Tensor},
+     {name="index", default=0}})
+
 wrap("tril",
      cname("tril"),
      {{name=Tensor, default=true, returned=true},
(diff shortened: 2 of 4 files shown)


lib/THC/THCTensorMath2.cu
```diff
@@ -660,3 +660,55 @@ void THCudaTensor_randn(THCState *state, THCudaTensor *r_, THLongStorage *size)
...
+  long p3 = start + 2;
+  if (p3 > nelem) return;
+  src_[p1] = x[p2]*y[p3] - x[p3]*y[p2];
+  src_[p2] = x[p3]*y[p1] - x[p1]*y[p3];
+  src_[p3] = x[p1]*y[p2] - x[p2]*y[p1];
+}
+
+THC_API void THCudaTensor_cross(THCState *state, THCudaTensor *self, THCudaTensor *x, THCudaTensor *y, int dimension)
+{
+  THAssert(THCudaTensor_checkGPU(state, 3, self, x, y));
```
apaszke: Please use `THCUNN_assertSameGPU` here. It does the same thing, but prints a nice error message if the check fails.

soumith:
cc: @wickedfoo @nicolasvasilache . The author of this PR is Zeming, please review when you get a chance.

lib/THC/THCTensorMath2.cu
```diff
@@ -660,3 +660,55 @@ void THCudaTensor_randn(THCState *state, THCudaTensor *r_, THLongStorage *size)
   THCudaTensor_resize(state, r_, size, NULL);
   THCudaTensor_normal(state, r_, 0, 1);
 }
+
+__global__ void THCudaTensor_kernel_cross(float *src_, float *x, float *y, long nelem) {
```
wickedfoo: use `TensorInfo<>` instead of `float*`

lib/THC/THCTensorMath2.cu
```diff
@@ -660,3 +660,55 @@ void THCudaTensor_randn(THCState *state, THCudaTensor *r_, THLongStorage *size)
...
+  THCudaTensor_resizeAs(state, self, x);
+  THCudaTensor *nx = THCudaTensor_newContiguous(state, x);
+  THCudaTensor *ny = THCudaTensor_newContiguous(state, y);
+  THCudaTensor_transpose(state, x, NULL, dimension, nd-1);
+  THCudaTensor_transpose(state, y, NULL, dimension, nd-1);
+
+  long threadsize = 192;
+  long MG = 65535;
+  dim3 threads(threadsize);
+  dim3 grid(min(MG, nelem), min(MG, nelem/MG + 1), min(MG, nelem/MG/MG));
```
wickedfoo: The grid is not taking into account `threadsize` at all; the kernel is over-launching threads. 
If the input tensor `x` has 12 elements, you will launch a grid with 192 threads x 4 blocks, whereas a single block will suffice.

Also, a construction like `nelem/MG +1` is overlaunching as well if nelem is an even multiple of MG. This is what the divup construction is for.

Fixing the number of threads you launch is ok, but the number of blocks that you launch should be related to `divup(nelem, (threadsize * 3))`.

Instead of doing this kind of math, it might make more sense to have the kernel itself loop in a grid-stride loop

https://devblogs.nvidia.com/parallelforall/cuda-pro-tip-write-flexible-kernels-grid-stride-loops/

Even better, I think you can actually use the THCTensorApply templates, you just need to transpose the size-3 dimension to innermost (which you are doing), but then set the size of that dimension to 1 in the TensorInfo; that will allow you to stride by 3. You could either ensure that the size-3 dimension is also contiguous and access the other elements by the points you are given, `p + 1` and `p + 2`, or use the stride for that dimension to get the other points.

Of course, the kernel will be fastest if the size-3 dimension is contiguous in memory. The transposes might still be desired, but the pattern with the rest of cutorch is that we use the THCTensorApply stuff to operate on non-contiguous tensors directly without newContiguous, to avoid cudaMalloc.

ebetica:
Gotcha, I didn't think of setting the size to 1 temporarily. If I do that I don't think I need to transpose them at all, and it looks in line with what torch does. I hope there isn't any weirdness that happens with async and setting the size back from 1 to 3, but the tests look like they're all passing.

lib/THC/THCTensorMath2.cu
```diff
@@ -660,3 +660,47 @@ void THCudaTensor_randn(THCState *state, THCudaTensor *r_, THLongStorage *size)
...
+
+  THArgCheck(dimension >= 0 && dimension < nd, 3, "dimension %d out of range", dimension+1);
+  THArgCheck(THCudaTensor_size(state, x, dimension) == 3, 3,
+      "dimension %d does not have size 3", dimension+1);
+  THCudaTensor_resizeAs(state, self, x);
+
+  long sx = THCudaTensor_stride(state, x, dimension);
+  long sy = THCudaTensor_stride(state, y, dimension);
+  long so = THCudaTensor_stride(state, self, dimension);
+  x->size[dimension] = 1;
```
wickedfoo: Don't modify data structures directly, use the (cu)Torch API (narrow, view, etc.) and create a new tensor that is a view onto the same tensor.

It is also possible that the read-only arguments here might be operated on in other threads as well, in which case you are modifying the same tensor that another thread is potentially operating on, too.

This creates a dependency on the underlying representation of a tensor, and bypasses the entire notion of having an API in the first place.

lib/THC/THCTensorMath2.cu
```diff
@@ -660,3 +660,47 @@ void THCudaTensor_randn(THCState *state, THCudaTensor *r_, THLongStorage *size)
   THCudaTensor_resize(state, r_, size, NULL);
   THCudaTensor_normal(state, r_, 0, 1);
 }
+
+struct TensorCrossOp {
+  TensorCrossOp(long sx, long sy, long so) : sx(sx), sy(sy), so(so) {}
+
+  __device__ __forceinline__ void operator()(float* out, float* x, float*y) {
+    *(out+0*so) = *(x+1*sx) * (*(y+2*sy)) - *(x+2*sx) * (*(y+1*sy));
+    *(out+1*so) = *(x+2*sx) * (*(y+0*sy)) - *(x+0*sx) * (*(y+2*sy));
```
wickedfoo: a nit:

```out[1 * so] = x[2 * sx] * y[0 * sy] - x[0 * sx] * y[2 * sy];```

etc. looks much more readable to me.

also, please float the binary operators in whitespace?

File: lib/THC/THCTensorMath2.cu [hunk 6]
```diff
@@ -660,3 +660,47 @@ void THCudaTensor_randn(THCState *state, THCudaTensor *r_, THLongStorage *size)
...
+
+THC_API void THCudaTensor_cross(THCState *state, THCudaTensor *self, THCudaTensor *x, THCudaTensor *y, int dimension)
+{
+  THAssert(THCudaTensor_checkGPU(state, 3, self, x, y));
+
+  int i;
+  long nd = THCudaTensor_nDimension(state, x);
+  long nelem = THCudaTensor_nElement(state, x);
+  THArgCheck(nd == THCudaTensor_nDimension(state, y), 1, "tensors must have same number of dimensions");
+  for (i=0; i<nd; i++) {
```
wickedfoo: nit: please float operators in whitespace for readability

```for (i = 0; i < nd; i++) {```

File: lib/THC/THCTensorMath2.cu [hunk 7]
```diff
@@ -660,3 +660,47 @@ void THCudaTensor_randn(THCState *state, THCudaTensor *r_, THLongStorage *size)
...
+    THArgCheck(THCudaTensor_size(state, x, i) == THCudaTensor_size(state, y, i), 1, "dimension %i of x and y does not match", i);
+    if (dimension < 0 && THCudaTensor_size(state, x, i) == 3) dimension = i;
```
wickedfoo: nit: please use braces for every if/else/elseif/for/while/do-while statement

wickedfoo:
looks good to me except for setting the underlying tensor data directly, please use the view/narrow/etc. functions instead to create a view of the tensor as needed, since this direct manipulation when used for the read-only arguments may conflict with actions in other threads.

also added some readability / cleanliness suggestions

ebetica:
Used the API and cleanliness suggestion like you suggested.

wickedfoo:
looks good to me

cc @soumith

PR #367, opened 2016-03-14
kernel p2p access and non-blocking streams

wickedfoo:
Adds a runtime option to allow within-kernel access to the memory space of other GPUs, if device-level p2p access is also enabled. By default, we still disallow direct access, but one can opt out of it (there's still a #define flag that can enable the opt-out, but @soumith told me to keep it in for a little bit).

I didn't add the flag for all pairs of GPUs like the p2p access option; that seems a little overkill to me personally.

Also added an option for `reserveStreams` that allows non-blocking streams to be created.

Updated documentation.


soumith:
Thanks!

Issue #339, opened 2016-02-23
sort() broken when total # elements is > 2^24 and sorting dimension > 2048 but <= 2^24


wickedfoo:
:(

Another victim of using float as an index.

The fallback Thrust sort doesn't work in this case, as there is a stage in the sorting where we have a global index.

Fix coming.

The real fix is to change the API to return a CudaLongTensor.

soumith:
@dominikgrewe are you guys okay with the change in the API (in a longer-term PR) to return a CudaLongTensor?

dominikgrewe:
I'm certainly in favour of moving to CudaLongTensor in the long term. As long as we keep things consistent with Torch it hopefully won't cause too many disruptions.

PR #369, opened 2016-03-14
fix large sorts

wickedfoo:
fixes https://github.com/torch/cutorch/issues/339

this is due to using a THCudaTensor to store global indices, which can go beyond 2^24 even if the sorting slice is small.

Changed it to use the index tensor as int before final conversion back to the per-slice index as float. Now the sort should work for all cases as long as the sorting slice itself is <= 2^24.

The real fix here is to change :sort() and :topk() to use and return a THCudaLongTensor as indices, but not going to do that in this diff (another one later on).


wickedfoo:
also removes some excess specializations for sort(), to cut down on compile time

Issue #366, opened 2016-03-13
Dead code in torch/generic/Tensor.c


apaszke:
[This piece](https://github.com/torch/cutorch/blob/master/torch/generic/Storage.c#L172-L189) is never generated, since only `THC_REAL_...` macros are defined. Also, it uses `memmove` on a CUDA pointer, so it's probably a remnant of a CPU implementation. Should I remove it or fix it?

soumith:
prob remove it for now.

PR #371, opened 2016-03-15
Remove dead storage code

apaszke:
It was guarded by TH_REAL... definitions, but only THC headers are used when compiling this file.


apaszke:
Fixes #366

PR #383, opened 2016-04-10
Remove use of broken THInf macro, which caused UB

tkoeppe:
See also https://github.com/torch/torch7/pull/634.

### DIFF
--- a/lib/THC/THCTensorMath.cu
+++ b/lib/THC/THCTensorMath.cu
@@ -7,6 +7,7 @@
 #include "THCReduceAll.cuh"
 
 #include <thrust/functional.h>
+#include <cfloat>
 
 struct TensorFillOp {
   TensorFillOp(float v) : val(v) {}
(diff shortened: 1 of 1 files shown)


soumith:
I merged https://github.com/torch/torch7/pull/634 , is this PR still needed? wont it now just default to FLT_MAX?

tkoeppe:
@soumith: No, all three PRs are needed. Because of  torch/torch7#634, there is now no global macro `THInf` anymore (it's only available during inclusions of `generic/` files), so a) this macro no longer exists for the present use case, but b) it would also be wrong: we don't *want* a type-speficic `THInf` here; we actually want the one for `float`.

Previously code like this would have caused UB on account of converting to `float` a value that cannot be represented.

hughperkins:
Hi guys @soumith  This issue broke also OpenCL Torch,  https://github.com/hughperkins/cltorch/pull/65  I dont suppose... could it be possible for me to subscribe to some specific repo or announcement list, or similar, so that I can find out about rapidly any breaking changes?  I know I could just put 'watch' on this repo, but you guys do so much work, so quickly, and the vast majority just works perfectly :-)

tkoeppe:
@hughperkins: I'm sorry about that. I was aware that I didn't test all users of Torch for this pretty invasive change. I just didn't see a better way of doing it. Thanks for your patience. I don't plan to propose any similarly invasive changes again.

PR #388, opened 2016-04-14
Extend README with new tensor types

bshillingford:
The README was a bit out of date. Added a listing of new tensor types including FP16, and `createCudaHostTensor`.


fmassa:
I think we should also add a notice saying that at the moment the only supported operations on these tensors are copy. What do you think ?

bshillingford:
Agreed, I've added that as a note to the intro and the section.

soumith:
Awesome! Thanks Brendan.

bshillingford:
Thanks!

PR #391, opened 2016-04-15
description that sort orders the 'key' tensor numerically

hughperkins:
This is sort of borderline, but I had to google around for 15-20 minutes to find this information, passing through:
- thrust doc: https://github.com/thrust/thrust/wiki/Quick-Start-Guide#sorting (which states that sort is based on key values, in alignment with behavior in standard library)
- c++ doc: http://www.cplusplus.com/reference/map/map/ (which states that default sort on a map is based on the key values)

Intuitively, I was expecting sort to be by value.

(Edit: sort works on cltorch now :-D )


soumith:
Thanks Hugh!

PR #394, opened 2016-04-19
few changes as required for cudnn fp16 support

lukasc-ch:
Minimal changes required to support fp16 from the cudnn package.


soumith:
Thanks :)

PR #399, opened 2016-04-30
Fix unchecked memory access when indexing a CudaTensor

htwaijry:
I ran into a segmentation fault when I accidentally tried to index a CudaTensor with an empty LongTensor.

To reproduce:
```
require 'cutorch'
a=torch.CudaTensor({1,2,3})
a:index(1, torch.LongTensor({}))
```

The invalid access happens in line 685 of  lib/THC/THCTensorIndex.cu, when getting size[0] of the indices tensor.

I added a THArgCheck before the access to check for correct dimensions.

-Hani


soumith:
Thanks Hani!

PR #405, opened 2016-05-13
cudaDeviceReset is not really deprecated

nicolasvasilache:
grep -Ri -3 deviceReset /usr/local/cuda-7.5/include | grep -3 deprec

… Its functionality is identical to the  non-deprecated function
::cudaDeviceReset(), which should be used instead

Reg. extern __host__ cudaError_t CUDARTAPI cudaThreadExit(void);


soumith:
cutorch.deviceReset was depreceated, not cuda's deviceReset. It was done because people were shooting themselves in the foot left and right. After the deviceReset is called, all existing cuda tensors are invalid, but cutorch keeps them alive as standard practice.

nicolasvasilache:
The lack of deviceReset can prevent nvprof from getting the proper information which is quite annoying. Shouldn't people that use cuda functions incorrectly fix their code? 
Would it be better to create a new cutorch function, say cutorch.deviceResetAtExit to make it more explicit and still keep backward compatibility?
I don't think limiting the ability to use nvprof is very satisfying..

wickedfoo:
@nicolasvasilache This deprecation was deliberate, since it is basically a nuclear option. It invalidates every device allocation, stream/event, etc, which are still referenced by live Torch objects, which we can't go and invalidate because we don't know what they are. Upon the next use, cudaErrors will be sent or worse a segfault will occur. I can't think of any use case for which it outside of a standalone C/C++ program where you know the universe of objects being used.

Is the issue not flushing perf info to nvprof? In order to fix that, add a `cudaDeviceSynchronize()` call before program exit.

wickedfoo:
(i.e., I don't know why this function was added to cutorch in the first place. I think it may have been from its earliest inception, where it looked like the CUDA API was just being copied into Lua)

nicolasvasilache:
Nope, cudaDeviceSynchronize() does not do it and nvprof is pretty explicit.

==2894542== Warning: Some profiling data are not recorded. Make sure cudaDeviceReset() is called before application exit to flush profile data.

I still think there is value in using this. I understand that people have used it improperly and it should be more explicit that it is invalidating everything.

Another way if you really don't want this in the cutorch API would be to register cudaDeviceReset with atexit.

nicolasvasilache:
will solve with the power of ffi

PR #409, opened 2016-05-18
stop cutorch clobbering cltorch, if `require`d after cltorch

hughperkins:
This isnt strictly necessary any more, since cltorch now runs from a stable torch distro.  But it might still be nice to add this sometime?

This patch makes the cutorch monkey-patch not overwrite any other module's existing monkey-patching on getmetatable(RealTensor).copy, in the case that `cutorch is`require`d after the other module.


hughperkins:
(Relevant doc https://github.com/hughperkins/cltorch#co-existence-with-cutorch )

soumith:
@torchbot test this

soumith:
@torchbot test this

soumith:
@torchbot test this

PR #412, opened 2016-05-19
Fix baddbmm for non-contiguous results tensor

fmassa:
Last changes required after https://github.com/torch/torch7/pull/672 , as discussed in https://github.com/torch/torch7/commit/8ae0a217f945606be817b22c6dd848ae9bb230d5#commitcomment-17542668
All tests passes in my machine.


soumith:
Thanks!

fmassa:
Thanks Soumith !

PR #415, opened 2016-05-26
Using new CMake FindCUDA elements

borisfom:
Hi Soumith,
I am getiing arch portion of FindCUDA into CMake, here:
https://github.com/Kitware/CMake/pull/237. 

With this change, I would finally be able to drop my cutorch clone as it would work with or without new bits of FindCUDA installed.


soumith:
Thanks!

PR #420, opened 2016-06-01
Using new CMake FindCUDA elements (#415)

borisfom:
Fixed the problem with previous PR (${} in IF() was evaluated which broke some cmake versions).
Improved "feature testing" for new function (as fp16 went into CMake with different PR, also it makes more sense).


soumith:
Thanks, tested this one, works.

soumith:
apparently, ${} is always dereferenced. 

http://stackoverflow.com/questions/15385252/cmake-defined-does-not-seems-to-recognize-variable

PR #429, opened 2016-06-11
Travis build tests

szagoruyko:
Only test builds for now with cuda-6.5


soumith:
there is no travis_cuda_install.sh ?

szagoruyko:
oy it's in distro, the path is incorrect. let me fix

PR #431, opened 2016-06-14
Reduction functions + masked* + BLAS work for data types

wickedfoo:
Converted reduction functions and masked* functions
Added Lua metatable support to all types for add/div/sub/unm/mul
Masked functions (and all/any) now require CudaByteTensor, like Torch7 requiring ByteTensor
More __index / __newindex work (so e.g., t[t:gt(0.5)] now works)


wickedfoo:
min(dim) and max(dim) return CudaLongTensor now as well

basically, anything index-ish I will convert to CudaLongTensor to be similar to the CPU version

wickedfoo:
updating diff with BLAS conversion shortly, would have sent it separately but it had some incompatibilities

wickedfoo:
Converted all BLAS functions to work on float + double. Ones for which there was a straightfoward half version (e.g., Hgemm) were converted too.

szagoruyko:
@wickedfoo this is awesome, I am testing your fork.
There is bit of an issue with compile time though, it takes very long time (especially if you have several arch cards in one machine), and `require 'cutorch'` takes 10s each time.

szagoruyko:
you forgot to add THCTensorMathBlas.h to CMakeLists.txt

```diff
diff --git a/lib/THC/CMakeLists.txt b/lib/THC/CMakeLists.txt
index de6e7aa..9f259ac 100644
--- a/lib/THC/CMakeLists.txt
+++ b/lib/THC/CMakeLists.txt
@@ -203,6 +203,7 @@ INSTALL(FILES
           generic/THCTensorMasked.cu
           generic/THCTensorMath.h
           generic/THCTensorMath.cu
+          generic/THCTensorMathBlas.h
           generic/THCTensorMathCompare.h
           generic/THCTensorMathCompare.cu
           generic/THCTensorMathCompareT.h
```

Another thing, `THCudaBlas_gemm` was changed and breaks cunn. Maybe we should keep THCudaBlas_gemm = THCudaBlas_Sgemm for compat.

wickedfoo:
@szagoruyko for the "requiring 10s" part, how many GPUs do you have?

It's been on my to-do list for a while to make GPU devices lazily initialize in Torch. Startup time goes away if you make fewer devices visible via CUDA_VISIBLE_DEVICES. I haven't looked into the problem in detail, but from things I've seen I think it is related to enabling p2p access for all pairs of devices.

re: compile time, the .cu files just need to be split up further so they can compile independently. Nvidia tells me that compile time has improved greatly in CUDA 8.0. If you don't have enough cores for parallel compile, then I think it will be slow, since it is effectively instantiating the same code as before, except for all datatypes instead of just float32.

szagoruyko:
@wickedfoo it takes 1.8, 4.5, 7.0 and 8.0 seconds for 1, 2, 3 and 4 cards, all Titan X. I switched to the same architecture in one machine and it's not that bad to compile, we can see how much it improves with CUDA 8.0 and then maybe split code into smaller chunks.

lib/THC/CMakeLists.txt
```diff
@@ -198,10 +199,18 @@ INSTALL(FILES
...
+          generic/THCTensorMathCompare.cu
+          generic/THCTensorMathCompareT.h
+          generic/THCTensorMathCompareT.cu
           generic/THCTensorMathPairwise.h
           generic/THCTensorMathPairwise.cu
           generic/THCTensorMathPointwise.h
           generic/THCTensorMathPointwise.cu
+          generic/THCTensorMathReduce.h
+          generic/THCTensorMathReduce.cu
           DESTINATION "${THC_INSTALL_INCLUDE_SUBDIR}/THC/generic")
```
borisfom: MathBlas files are missing from the export, making CUNN fail.
Here : https://github.com/borisfom/cutorch/commit/ccff07f0616e1b5b5767eeb8b01e659f530c0ac9

lib/THC/THCBlas.cu
```diff
@@ -1,109 +1,79 @@
...
   }
-  THError("Cublas_dot only supports n, incx and incy "
-          "upto signed integer limits: %d", INT_MAX);
-  return -1;
+  THError("Cublas_Sgemv only supports m, n, lda, incx, incy"
+          "in the range 0 < [val] <= %d", INT_MAX);
 }
 
-/* Level 2 */
-void THCudaBlas_gemv(THCState *state, char trans, long m, long n, float alpha, float *a, long lda, float *x, long incx, float beta, float *y, long incy)
```
borisfom: Is it OK to not provide backward compatibility here ?  CUNN is failing.

lib/THC/THCBlas.cu
```diff
@@ -211,17 +202,75 @@ void THCudaBlas_gemm(THCState *state, char transa, char transb, long m, long n,
...
+    // Check for native Hgemm support
+    if (THC_nativeHalfInstructions(state)) {
+      THCublasCheck(cublasHgemm(THCState_getCurrentBlasHandle(state), opa, opb, i_m, i_n, i_k, &alpha, a, i_lda, b, i_ldb, &beta, c, i_ldc));
+    } else {
+      // Simulated Hgemm
+      float fAlpha = THC_half2float(alpha);
+      float fBeta = THC_half2float(beta);
+
+      THCublasCheck(cublasSgemmEx(THCState_getCurrentBlasHandle(state), opa, opb, i_m, i_n, i_k, &fAlpha,
+                                  a, CUBLAS_DATA_HALF, i_lda, b, CUBLAS_DATA_HALF, i_ldb, &fBeta, c, CUBLAS_DATA_HALF, i_ldc));
```
borisfom: This should be changed for CUDA 8.0. I fixed it with #ifdef here : https://github.com/borisfom/cutorch/commit/ccff07f0616e1b5b5767eeb8b01e659f530c0ac9

soumith:
this PR has now been pushed in via #456

PR #433, opened 2016-06-21
Fix torch.clamp behavior with CudaTensor.

### LINKED ISSUE #432: torch.clamp() changes input data in cuda
torch.clamp(a, min_val, max_val) changes the value of a in cuda. However, it doesn't change the value of a if using cpu instead of gpu.
For example
```lua
require 'cutorch'
require 'cunn'
local data = torch.DoubleTensor {1,2,3,4,5,6,7,8,9,10}:cuda()
local clamped = torch.clamp(data, 2, 7)
print(data)
```
the output will be 
2 2 3 4 5 6 7 7 7 7

luketwitter:
torch.clamp mistakely overwrites the tensor parameter instead of returning a new tensor when called with 3 arguments (e.g. torch.clamp(a, 1, 7)), when the tensor is a CudaTensor

Fixes #432

### DIFF
--- a/TensorMath.lua
+++ b/TensorMath.lua
@@ -1011,7 +1011,7 @@ wrap("multinomial",
 wrap("clamp",
      cname("clamp"),
      {{name=Tensor, default=true, returned=true, method={default='nil'}},
-      {name=Tensor, default=1},
+      {name=Tensor, method={default=1}},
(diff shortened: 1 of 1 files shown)


soumith:
Thank you :)

PR #439, opened 2016-06-29
added field driverVersion to cutorch

lukasc-ch:
Exposes the gpu driver version to Torch. This will be used in soumith/cudnn.torch#212.


soumith:
Thanks!

PR #453, opened 2016-07-29
1.0

soumith:
Cutting a version for the first time in history!

The changes in API are super minor, but just acting like a grown up these days...

cc: @jonathantompson


jonathantompson:
Awesome! Celebratory beers all around :-)

PR #456, opened 2016-07-29
reduce and BLAS work

soumith:
This commit by @wickedfoo (see https://github.com/torch/cutorch/pull/431 ) introduces breaking API changes.
Users who dont want to change their code can go to cutorch / cunn version 1.0-0 via:

luarocks install cunn 1.0-0


soumith:
Also see: https://github.com/torch/cutorch#api-changes-and-versioning

This PR btw incorporates changes from @borisfom

PR #457, opened 2016-08-02
Pass CMAKE_CXX_COMPILER to new select_compute_arch.cmake

szagoruyko:
Support passing "CC=gcc-4.8 CXX=g++-4.8" for autodetect in Ubuntu 16.04, otherwise it fails and tries to compile for all archs


soumith:
Thanks!

PR #459, opened 2016-08-07
scatter / gather to new types

soumith:
Also refactored a few tests to start making things easy to check across all types in both LongIndex mode and the old Index mode.

cc: @wickedfoo


lib/THC/THCTensorTopK.cu
```diff
@@ -531,8 +532,15 @@ THC_API void THCudaTensor_topk(THCState* state,
 
       THCudaTensor* sortedTopKIndices = THCudaTensor_new(state);
 
+      THCudaLongTensor *sortedIndicesLong = THCudaLongTensor_new(state);
```
soumith: this'll be fixed when i convert topK to multiple types.

PR #461, opened 2016-08-11
Index* for new types

soumith:
This makes the index* methods to use CudaLongTensor for indices.
Adds index* for all CUDA types (except indexAdd because need to implement atomicAdd for non float / int types)

Also changes the sort and topK API accordingly (but doesn't add new types in them yet)

Checklist before merging:

- [x] check and issue fixes to nn / cunn
- [x] check rnn
- [x] check dpnn
- [x] check inn
- [x] check cunnx (fails but code is half fixed and half isn't and mess. experimental, so ignore)
- [x] check that densecap works
- [x] fix half tests


soumith:
cc: @wickedfoo

lib/THC/THCTensorSort.cu
```diff
@@ -40,22 +40,23 @@ fillSliceWithIndex(TensorInfo<float, IndexType> out,
...
   const unsigned long offset =
-    IndexToOffset<float, IndexType, Dim>::get(slice, out);
-  float* base = &out.data[offset];
+    IndexToOffset<long, IndexType, Dim>::get(slice, out);
+  long* base = &out.data[offset];
 
   for (long i = threadIdx.x; i < sliceSize; i += blockDim.x) {
     // Torch indices are 1-based (hence the +1)
-    base[i * sliceStride] = (float) i + 1.0f;
+    base[i * sliceStride] = (long) i + 1;
```
wickedfoo: you can remove the (long), it's already long

test/test.lua
```diff
@@ -49,7 +49,7 @@ end
 
 local function checkHalf()
    if cutorch.hasHalf then
-      table.insert(typenames, 'CudaHalfTensor')
+       -- table.insert(typenames, 'CudaHalfTensor')
```
wickedfoo: what's the story here?

lib/THC/THCTensorTopK.cu
```diff
@@ -261,7 +261,7 @@ __global__ void gatherTopK(TensorInfo<float, IndexType> input,
                            IndexType numTopKSlices,
                            IndexType topKWithinSliceStride,
 
-                           TensorInfo<float, IndexType> indices,
+                           TensorInfo<long, IndexType> indices,
```
wickedfoo: there's a lot more that needs changing in this function to use long indices, if you are calling top-k on a tensor with more than `2^31 - 1` elements FYI

wickedfoo:
I think it looks good to me, besides the topk issues (it won't work for some tensors with more than 2^31 - 1 elements).

should add an assertion to topk that you aren't calling it on a tensor where the slice size that you are top-k-ing is > 2^31 - 1 elements.

(also sorry, github notifications have been going to my spam gmail for the past couple of months, hopefully i'll get them now)

test/test.lua
soumith: this needs to be fixed. i'll fix the tests before checking in.

soumith:
fixes https://github.com/torch/cutorch/issues/463

soumith:
fixes https://github.com/jcjohnson/densecap/issues/26 and https://github.com/jcjohnson/densecap/issues/22

PR #481, opened 2016-08-23
CNMEM pool integration

borisfom:
@colesbury @soumith : this is functionally complete and tested on torch tests and convnet/image-winners (no big difference there as no reallocations during iterations, please recommend decent benchmark to check performance).
May be used as default (with a smaller value or explicitly manipulated from Lua (not really for user, but may be very useful for CUDNN and such).
Implemented as per discusson in https://github.com/torch/cutorch/pull/443. 
CNMEM repo is included via subtree.
Designed to be extendable to try other kinds of pools (e.g CUB).

### DIFF
--- a/TensorMath.lua
+++ b/TensorMath.lua
@@ -724,6 +724,44 @@ for k, Tensor_ in pairs(handledTypenames) do
              {name="boolean", default=0}}
     )
 
+    wrap("squeeze",
+         cname("squeeze"),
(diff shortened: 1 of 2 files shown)


colesbury:
@borisfom, I think this is missing the files THCMemory.h and THCMemory.c

borisfom:
@colesbury : added missing files.

PR #501, opened 2016-09-12
README: CudaTensor doesn't support :diag()

CDLuminate:


soumith:
They were added to cutorch on July 27th. https://github.com/torch/cutorch/pull/451

CDLuminate:
Thanks ... I need to update the torch instance on my workstation.

PR #506, opened 2016-09-16
Add support for CUDA indexAdd

gchanan:
Adds indexAdd via atomicAdd for unsigned char, char, short, long,
half, double.  Integer types are templatized based on sizeof.
Floating point types are implemented via intrinsics.

### DIFF
--- a/lib/THC/THCTensorIndex.cu
+++ b/lib/THC/THCTensorIndex.cu
@@ -4,11 +4,12 @@
 #include "THCBlas.h"
 #include "THCTensorCopy.h"
 #include "THCTensorRandom.h"
+#include "THCHalf.h"
 #include "THCApply.cuh"
 #include "THCReduce.cuh"
 #include "THCDeviceUtils.cuh"
+#include "THCNumerics.cuh"
 #include <algorithm> // for std::min
-#include <cuda_fp16.h>
 
 // We prefer this kernel to avoid reloading index points if the number
 // of indices is a small number.
@@ -103,15 +104,16 @@ struct AtomicAddIntegerImpl<T, 1> {
   __device__ void operator()(T *address, T val) {
     unsigned int * address_as_ui =
         (unsigned int *) (address - ((size_t)address & 3));
-    unsigned int old = *address_as_ui, sum, newval, assumed;
-    unsigned int selectors[] = {0x3214, 0x3240, 0x3410, 0x4210};
-    unsigned int sel = selectors[(size_t)address & 3];
+    unsigned int old = *address_as_ui;
+    unsigned int shift = (((size_t)address & 3) * 8);
+    unsigned int sum;
+    unsigned int assumed;
 
     do {
       assumed = old;
-      sum = val +  (T)__byte_perm(old, 0, (size_t)address & 3 | 0x4440);
-      newval = __byte_perm(old, sum, sel);
-      old = atomicCAS(address_as_ui, assumed, newval);
+      sum = val + T((old >> shift) & 0xff);
+      old = (old & ~(0x000000ff << shift)) | (sum << shift);
+      old = atomicCAS(address_as_ui, assumed, old);
     } while (assumed != old);
   }
 };
@@ -121,12 +123,15 @@ struct AtomicAddIntegerImpl<T, 2> {
   __device__ void operator()(T *address, T val) {
     unsigned int * address_as_ui =
         (unsigned int *) ((char *)address - ((size_t)address & 2));
-    unsigned int old = *address_as_ui, sum, newval, assumed;
+    unsigned int old = *address_as_ui;
+    unsigned int sum;
+    unsigned int newval;
+    unsigned int assumed;
 
     do {
       assumed = old;
-      sum = val +  (T)__byte_perm(old, 0, ((size_t)address & 2) ? 0x4432 : 0x4410);
-      newval = __byte_perm(old, sum, ((size_t)address & 2) ? 0x5410 : 0x3254);
+      sum = val + (size_t)address & 2 ? T(old >> 16) : T(old & 0xffff);
+      newval = (size_t)address & 2 ? (old & 0xffff) | (sum << 16) : (old & 0xffff0000) | sum;
       old = atomicCAS(address_as_ui, assumed, newval);
     } while (assumed != old);
   }
@@ -136,7 +141,9 @@ template<typename T>
 struct AtomicAddIntegerImpl<T, 4> {
   __device__ void operator()(T *address, T val) {
     unsigned int * address_as_ui = (unsigned int *) (address);
-    unsigned int old = *address_as_ui, newval, assumed;
+    unsigned int old = *address_as_ui;
+    unsigned int newval;
+    unsigned int assumed;
 
     do {
       assumed = old;
@@ -150,7 +157,9 @@ template<typename T>
 struct AtomicAddIntegerImpl<T, 8> {
   __device__ void operator()(T *address, T val) {
     unsigned long long * address_as_ui = (unsigned long long *) (address);
-    unsigned long long old = *address_as_ui, newval, assumed;
+    unsigned long long old = *address_as_ui;
+    unsigned long long newval;
+    unsigned long long assumed;
 
     do {
       assumed = old;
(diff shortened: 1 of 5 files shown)


lib/THC/THCTensorIndex.cu
```diff
@@ -8,6 +8,7 @@
 #include "THCReduce.cuh"
 #include "THCDeviceUtils.cuh"
 #include <algorithm> // for std::min
+#include <cuda_fp16.h>
```
wickedfoo: include `THCHalf.cuh` instead, since cuda_fp16.h may not be available

lib/THC/THCTensorIndex.cu
```diff
@@ -94,6 +95,124 @@ __global__ void indexCopyLargeIndex(TensorInfo<T, IndexType> dst,
...
 
+template <typename T, size_t n>
+struct AtomicAddIntegerImpl;
+
+template<typename T>
+struct AtomicAddIntegerImpl<T, 1> {
+  __device__ void operator()(T *address, T val) {
+    unsigned int * address_as_ui =
+        (unsigned int *) (address - ((size_t)address & 3));
+    unsigned int old = *address_as_ui, sum, newval, assumed;
```
wickedfoo: Please don't use comma lists for declaration, it is prone to confusion generally speaking. Especially here, since there is an initialization only for one variable along with declaration without initialization for others.

```
unsigned int old = *address_as_ui;
unsigned int sum;
unsigned int newval;
unsigned int assumed;
```

lib/THC/THCTensorIndex.cu
```diff
@@ -94,6 +95,124 @@ __global__ void indexCopyLargeIndex(TensorInfo<T, IndexType> dst,
...
+
+__device__ void atomicAdd(short *address, short val) {
+  AtomicAddIntegerImpl<short, sizeof(short)>()(address, val);
+}
+
+__device__ void atomicAdd(long *address, long val) {
+  AtomicAddIntegerImpl<long, sizeof(long)>()(address, val);
+}
+
+#ifdef CUDA_HALF_INSTRUCTIONS
```
wickedfoo: This needs to be guarded by whether or not `half` is generally enabled, with `THCHalf.cuh`'s `CUDA_HALF_TENSOR`

lib/THC/THCTensorIndex.cu
```diff
@@ -94,6 +95,124 @@ __global__ void indexCopyLargeIndex(TensorInfo<T, IndexType> dst,
...
+#define HALF_ADD(a, b) __hadd(a, b)
+#else
+#define HALF_ADD(a, b) __float2half(__half2float(a) + __half2float(b))
+#endif
+
+__device__ void atomicAdd(half *address, half val) {
+  unsigned int * address_as_ui =
+      (unsigned int *) ((char *)address - ((size_t)address & 2));
+  unsigned int old = *address_as_ui, assumed;
```
wickedfoo: same here re: declaration

lib/THC/THCTensorIndex.cu
```diff
@@ -94,6 +95,124 @@ __global__ void indexCopyLargeIndex(TensorInfo<T, IndexType> dst,
...
+    hsum = HALF_ADD(val, *hp);
+    memcpy(hp, &hsum, sizeof(half));
+    old = atomicCAS(address_as_ui, assumed, old);
+   } while (assumed != old);
+}
+
+// from CUDA C Programmic Guide
+__device__  void atomicAdd(double *address, double val) {
+  unsigned long long int* address_as_ull = (unsigned long long int*)address;
+  unsigned long long int old = *address_as_ull, assumed;
```
wickedfoo: same here re: declaration

lib/THC/THCTensorIndex.cu
```diff
@@ -94,6 +95,124 @@ __global__ void indexCopyLargeIndex(TensorInfo<T, IndexType> dst,
...
+    } while (assumed != old);
+  }
+};
+
+template<typename T>
+struct AtomicAddIntegerImpl<T, 2> {
+  __device__ void operator()(T *address, T val) {
+    unsigned int * address_as_ui =
+        (unsigned int *) ((char *)address - ((size_t)address & 2));
+    unsigned int old = *address_as_ui, sum, newval, assumed;
```
wickedfoo: same here re: initialization

lib/THC/THCTensorIndex.cu
```diff
@@ -94,6 +95,124 @@ __global__ void indexCopyLargeIndex(TensorInfo<T, IndexType> dst,
...
+      old = atomicCAS(address_as_ui, assumed, newval);
+    } while (assumed != old);
+  }
+};
+
+template<typename T>
+struct AtomicAddIntegerImpl<T, 4> {
+  __device__ void operator()(T *address, T val) {
+    unsigned int * address_as_ui = (unsigned int *) (address);
+    unsigned int old = *address_as_ui, newval, assumed;
```
wickedfoo: same here re: initialization

lib/THC/THCTensorIndex.cu
```diff
@@ -94,6 +95,124 @@ __global__ void indexCopyLargeIndex(TensorInfo<T, IndexType> dst,
...
+      old = atomicCAS(address_as_ui, assumed, newval);
+    } while (assumed != old);
+  }
+};
+
+template<typename T>
+struct AtomicAddIntegerImpl<T, 8> {
+  __device__ void operator()(T *address, T val) {
+    unsigned long long * address_as_ui = (unsigned long long *) (address);
+    unsigned long long old = *address_as_ui, newval, assumed;
```
wickedfoo: same here re: initialization

lib/THC/THCTensorIndex.cu
```diff
@@ -94,6 +95,124 @@ __global__ void indexCopyLargeIndex(TensorInfo<T, IndexType> dst,
...
+__device__ void atomicAdd(short *address, short val) {
+  AtomicAddIntegerImpl<short, sizeof(short)>()(address, val);
+}
+
+__device__ void atomicAdd(long *address, long val) {
+  AtomicAddIntegerImpl<long, sizeof(long)>()(address, val);
+}
+
+#ifdef CUDA_HALF_INSTRUCTIONS
+#define HALF_ADD(a, b) __hadd(a, b)
```
wickedfoo: `HALF_ADD` is already effectively defined by `THCNumerics<half>::add(a, b)`

lib/THC/THCTensorIndex.cu
```diff
@@ -94,6 +95,124 @@ __global__ void indexCopyLargeIndex(TensorInfo<T, IndexType> dst,
...
+__device__ void atomicAdd(half *address, half val) {
+  unsigned int * address_as_ui =
+      (unsigned int *) ((char *)address - ((size_t)address & 2));
+  unsigned int old = *address_as_ui, assumed;
+  half hsum;
+
+  do {
+    // byte_perm doesn't work with halfs, let's manually memcpy
+    assumed = old;
+    half *hp = (half *)((char *)(&old) + ((size_t)address & 2));
```
wickedfoo: I don't have high hopes here that the compiler is able to statically determine that lines 194-196 is equivalent to bit manipulation. Taking the address of a value in a register and performing this manipulation is liable to move the value from being in a GPU register and will force it to local memory, which is slow.

Since there are only 2 possible offset states (first 16-bit word or second 16-bit word), I would do something like this:

```
half hsum;
// I think GPUs are little endian?
hsum.x = address & 2 == 0 ? (old & 0xffff) : (old >> 16);
hsum = THCNumerics<half>::add(hsum, val);
// write the 16-bit word back into `old` and then do the CAS stuff
```

predication on the GPU should make the `? :` pretty efficient, there are only two states here. You could avoid the predication as well by determining the shift required by inspecting `address`.

lib/THC/THCTensorIndex.cu
```diff
@@ -94,6 +95,124 @@ __global__ void indexCopyLargeIndex(TensorInfo<T, IndexType> dst,
...
+  __device__ void operator()(T *address, T val) {
+    unsigned int * address_as_ui =
+        (unsigned int *) (address - ((size_t)address & 3));
+    unsigned int old = *address_as_ui, sum, newval, assumed;
+    unsigned int selectors[] = {0x3214, 0x3240, 0x3410, 0x4210};
+    unsigned int sel = selectors[(size_t)address & 3];
+
+    do {
+      assumed = old;
+      sum = val +  (T)__byte_perm(old, 0, (size_t)address & 3 | 0x4440);
```
wickedfoo: what about overflow? doesn't seem correct to me since you could be bleeding into other bytes?

wickedfoo: the code might be easier to understand with manual shifting and masking as well, would make it easier to deal with overflow by restricting the add to `T` anyways.

It might be faster as well, since line 108 is probably forcing usage of local memory.

Generally speaking, in most kernels there is no actual linear region of memory that is a "stack" (except something where the recursion is indeterminate) and the GPU tries very hard to compile code using register assignment. Dynamic indexing into local vars/arrays/pointer math into addresses of local vars will force things to go into local memory.

lib/THC/THCTensorIndex.cu
```diff
@@ -94,6 +95,124 @@ __global__ void indexCopyLargeIndex(TensorInfo<T, IndexType> dst,
...
+  AtomicAddIntegerImpl<long, sizeof(long)>()(address, val);
+}
+
+#ifdef CUDA_HALF_INSTRUCTIONS
+#define HALF_ADD(a, b) __hadd(a, b)
+#else
+#define HALF_ADD(a, b) __float2half(__half2float(a) + __half2float(b))
+#endif
+
+__device__ void atomicAdd(half *address, half val) {
```
wickedfoo: also needs to be guarded with `#ifdef CUDA_HALF_TENSOR`

wickedfoo:
ah I guess the __byte_perm is selecting from two different values, so there aren't overflow issues. Still, it's confusing.

lib/THC/THCTensorIndex.cu
wickedfoo: nevermind, i guess there isn't overflow because you're selecting again below to re-permute, but i didn't stare at the selection indices long enough to determine if it was correct. I think just shifting + masking the 8 or 16 bit word from the 32 bit word that you want and then putting it back is easier to follow, and then you aren't using an array of selector values either based on offset (which might go into local memory or involve indexing in constant memory)

lib/THC/THCTensorIndex.cu
gchanan: I'm not sure what you mean here -- this needs to be guarded by CUDA_HALF_TENSOR?  The other cases I see CUDA_HALF_TENSOR used are with the half data type, but this should only be templatized for integer types.  Or does this need to be guarded by something else? (CUDA_ARCH versions for the atomicCAS functions?)

lib/THC/THCTensorIndex.cu
gchanan: ah, thanks for the tip on local memory.  I originally had something close to the above, but didn't want to deal with endian-ness.  I'll investigate.

lib/THC/THCTensorIndex.cu
wickedfoo: I mean the multiple declarations in one line here; split them out to their own declarations (a style thing).
```
unsigned int old = *address_as_ui;
unsigned int sum;
...
```

soumith:
@wickedfoo greg updated the PR with the fixes. wdyt?

soumith:
Thanks Greg!

PR #510, opened 2016-09-22
Add CUDA caching allocator

colesbury:
The allocator can be enabled by setting the environment variable `THC_CACHING_ALLOCATOR=1`

cc @wickedfoo

### DIFF
--- a/init.c
+++ b/init.c
@@ -1,6 +1,7 @@
 #include "utils.h"
 #include "luaT.h"
 #include "THCGeneral.h"
+#include "THCCachingAllocator.h"
 #include "THCTensorRandom.h"
 #include "THCHalf.h" // for CUDA_HALF_TENSOR
 
@@ -955,6 +956,13 @@ int luaopen_libcutorch(lua_State *L)
   luaL_setfuncs(L, cutorch_stuff__, 0);
 
   THCState* state = (THCState*)malloc(sizeof(THCState));
+  memset(state, 0, sizeof(THCState));
+
+  char* thc_caching_allocator = getenv("THC_CACHING_ALLOCATOR");
+  if (thc_caching_allocator && strcmp(thc_caching_allocator, "1") == 0) {
+    THCCachingAllocator_init(&state->cudaDeviceAllocator);
+  }
+
   THCudaInit(state);
 
   /* Register torch.CudaHostAllocator. */
@@ -1028,7 +1036,7 @@ int luaopen_libcutorch(lua_State *L)
   int driverVersion;
   THCudaCheck(cudaDriverGetVersion(&driverVersion));
   lua_pushinteger(L, driverVersion);
-  lua_setfield(L, -2, "driverVersion");  
+  lua_setfield(L, -2, "driverVersion");
 
   /* when cutorch goes out of scope, we need to make sure THCState is properly
      shut down (so that memory doesn not leak. Since _state is a lightuserdata
--- a/lib/THC/THCCachingAllocator.cpp
+++ b/lib/THC/THCCachingAllocator.cpp
@@ -1,0 +1,291 @@
+#include "THCCachingAllocator.h"
+
+#include <cuda_runtime_api.h>
+#include <map>
+#include <memory>
+#include <mutex>
+#include <set>
+#include <unordered_map>
+
+//
+// Yet another caching allocator for CUDA device allocations.
+//
+// - Allocations are associated with a stream. Once freed, blocks can be
+//   re-allocated on the same stream, but not on any other stream.
+// - The allocator attempts to find the smallest cached block that will fit the
+//   requested size. If the block is larger than the requested size, it may be
+//   split. If no block is found, the allocator will delegate to cudaMalloc.
+// - If the cudaMalloc fails, the allocator will free all cached blocks that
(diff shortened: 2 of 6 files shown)


init.c
```diff
@@ -955,6 +956,13 @@ int luaopen_libcutorch(lua_State *L)
   luaL_setfuncs(L, cutorch_stuff__, 0);
 
   THCState* state = (THCState*)malloc(sizeof(THCState));
+  memset(state, 0, sizeof(THCState));
```
colesbury: THCudaInit now expected the caller to zero-initialize the THCState. This is so that the caller can set the `cudaDeviceAllocator` early (**before** any THCudaMalloc calls).

lib/THC/THCCachingAllocator.cpp
```diff
@@ -0,0 +1,282 @@
...
+    return a->size < b->size;
+  }
+  return a->ptr < b->ptr;
+}
+
+struct THCCachingAllocator
+{
+
+  typedef bool (*Comparison)(const Block*, const Block*);
+  typedef std::multiset<Block*, Comparison> FreeBlocks;
```
colesbury: Yeah, this should be set

lib/THC/THCCachingAllocator.cpp
```diff
@@ -0,0 +1,282 @@
...
+    } else {
+      size += 131072 - 1 - (size - 1) % 131072;
+    }
+    return size;
+  }
+
+  cudaError_t cuda_malloc_retry(int device, void** devPtr, size_t size)
+  {
+    cudaError_t err = cudaMalloc(devPtr, size);
+    if (err != cudaSuccess) {
```
borisfom: Yes calling cudaGetLastError() is crucial

lib/THC/THCCachingAllocator.cpp
```diff
@@ -0,0 +1,282 @@
...
+  // allocated blocks by device pointer
+  std::unordered_map<uintptr_t, Block*> allocated_blocks;
+
+  THCCachingAllocator() :
+      large_blocks(BlockComparator),
+      small_blocks(BlockComparator) {}
+
+  cudaError_t malloc(void** devPtr, size_t size, cudaStream_t stream)
+  {
+    std::lock_guard<std::mutex> lock(mutex);
```
borisfom: I think C++11 is now mainstream and should be used. 
Side comment: when last time I asked about using C++ in Torch at all (I guess, besides .cu as this is already in use) , the answer was a strict no. @soumith, did that change?

lib/THC/THCCachingAllocator.cpp
```diff
@@ -0,0 +1,282 @@
...
+  if (a->stream != b->stream) {
+    return (uintptr_t)a->stream < (uintptr_t)b->stream;
+  }
+  if (a->size != b->size) {
+    return a->size < b->size;
+  }
+  return a->ptr < b->ptr;
+}
+
+struct THCCachingAllocator
```
colesbury: After discussing with @wickedfoo, I've moved THCCachingAllocator outside the anonymous namespace, but did not add it to the header file since it contains C++

init.c
```diff
@@ -955,6 +956,13 @@ int luaopen_libcutorch(lua_State *L)
...
+
+  char* thc_caching_allocator = getenv("THC_CACHING_ALLOCATOR");
```
colesbury: This is specific to the Lua API. Other interfaces to THC may want to always use the caching allocator (or swap the defaults so that it's enabled by default and requires the env variable to disable it).

colesbury:
Regarding C++11, I've tested this with GCC 4.4.7 which has pretty limited C++11 support. If we don't use any C++11 here, the problem is that we'll have to do platform-specific mutexes (pthread and windows).

soumith:
Woohoo! Can you add to the README how to enable this in cutorch?

anibali:
I believe the use of C++11 in this PR has stopped the Torch distro from building properly for me. This is what I see in my output:

```
In file included from /usr/include/c++/4.8/mutex:35:0,
                 from /root/torch/extra/cutorch/lib/THC/THCCachingAllocator.cpp:6:
/usr/include/c++/4.8/bits/c++0x_warning.h:32:2: error: #error This file requires compiler and library support for the ISO C++ 2011 standard. This support is currently experimental, and must be enabled with the -std=c++11 or -std=gnu++11 compiler options.
 #error This file requires compiler and library support for the \
  ^
```

I'm guessing that the use of C++11 is now going to trip up a bunch of people depending on which compiler version they use (I'm using `gcc (Ubuntu 4.8.4-2ubuntu1~14.04.3) 4.8.4`).

My personal opinion is similar to @wickedfoo in that I believe staying away from C++11 would be nice. However, failing that, it might be necessary to add more compiler flags to the build process?

xternalz:
@anibali try `SET(CMAKE_CXX_FLAGS "-std=c++11 ${CMAKE_CXX_FLAGS}")`

soumith:
@anibali fixed now via https://github.com/torch/cutorch/commit/684659238eab7ed3a817fcf04099c8393ea68f60

anibali:
@soumith I just gave your fix a try and it works like a charm. Thanks a lot!

PR #516, opened 2016-09-28
Add support for GenerateFloatTypes, for use with cunn.

gchanan:
Add support for GenerateFloatTypes, for use with cunn.


soumith:
Thanks Greg!

PR #517, opened 2016-09-29
Move more functions into generic

killeent:
This PR moves the following Tensor functions into the generic directory:

- mean/meanall are extended to all tensor types
- clamp to all tensor types
- cross to all tensor types
- cmin/cmax to all tensor types
- std/stdall to float/half/double tensors
- var/varall to float/half/double tensors
- norm/normall to float/half/double tensors
- renorm to float/half/double tensors
- lerp to float/half/double tensors

Required refactoring a lot of float specific code into .cuh files. 

Test Plan:

- updated unit tests to compare using new tensor types


killeent:
Moving into #530

PR #518, opened 2016-09-29
Compile fixes for picky compilers / stl versions

rsfbarreira:


soumith:
thanks. for information, can you tell me what compiler you used, which made this picky

Issue #520, opened 2016-09-29
We should temporarily force pseudo-fp16 mode (HAS_HALF_INSTRUCTIONS=FALSE) for Pascal


borisfom:
@soumith, @szagoruyko : 
We should temporarily force "pseudo-fp16" mode (32-bit math for 16-bit floats, HAS_HALF_INSTRUCTIONS=FALSE switch) for Pascal.
Background of this issue is desrcibed in https://github.com/torch/cutorch/issues/515.
I believe, until both that and CUDNN5 fp16 crashes issues are fixed (which may as well linger until CUDNN6 timeframe), cutorch should not enable true fp16 math by default to prevent Pascal users from negative experience (code working fine on Maxwell would work slower on Titan XP and/or crash on P100. 
Before  the solution to https://github.com/torch/cutorch/issues/515 is in place, pseudo-fp16 has to be compile-time.
After that and https://github.com/torch/cunn/issues/315 set in place, it should be runtime but still off by default (or on for GP100 only).
I will open corresponding issue in torch.cudnn with cudnn specific details.

PR #521, opened 2016-09-30
temporarily force pseudo-fp16 mode (HAS_HALF_INSTRUCTIONS=FALSE) for …

borisfom:
…Pascal: as per my suggestion in https://github.com/torch/cutorch/issues/520


borisfom:
@soumith: FYI, next NVIDIA Torch package will be shipping with this change.

soumith:
we really prefer that you guys do not fork torch. i am not sure why you are going that route.

soumith:
and for a proper fix so that P100 users aren't left in the dust, please see discussion here: https://github.com/torch/cutorch/issues/544

borisfom:
@soumith: we needed this PR for the release, and it was not integrated so we had to use the fork.

PR #528, opened 2016-10-07
bug fix for wrong usage of checkGPU && port to windows with msvc

BTNC:
Hi,

This PR contains: (1) fix wrong usage of checkGPU in indexCopy, indexAdd; (2) work on windows with msvc.

All tests under test/ passed except some random failures like logNormal etc. Those kinds of failures are due to poor test designs and should be fine.

By the way, it is clear cutorch is using long as 64 bit integer, some evidences can be found from this commit. Could you shed some light if it is ok to replace long with int64_t or similar?  Besides, long long is used a lot in cutorch in order to represent numbers that could be bigger than long, while long is assumed to be 64 bit. On the contrary, there are some long long to long conversions. In short, there are so many integer type mismatches, is there a plan to clear them all?

Thanks,


soumith:
I think all of TH and THC needed to be cleaned up of long, and instead be switched to fixed width types like int32_t and int64_t. It's historical cruft, unfortunately.

> Could you shed some light if it is ok to replace long with int64_t or similar?

Definitely.

> On the contrary, there are some long long to long conversions. In short, there are so many integer type mismatches, is there a plan to clear them all?

I think we dont notice them because it works fine in 64-bit *nix systems, under which 90% of us are.

Issue #515, opened 2016-09-28
cutorch should provide an option to force float math on half values (pseudo-fp16)


borisfom:
It proved to be beneficial to be able to force using 32-bit float math on 16-bit float data:
1. That's how all math in Maxwell chips work. 
2. Currently, cutorch forces use of true fp16 math on Pascal - while in fact it's not beneficial at all on non-P100 cards (slower), and risky even on P100 (cudnn still has some bugs around it, as we know).
   So, the use case where Pascal user would want to override math for 'half' to be 32-bit 'float' is very realistic - in fact, a necessity. 
   Since Torch is an interpreter, it can also be done without runtime cost (and even can make cutorch code cleaner - now it's ridden with ifdefs for half).

I am opening it as an issue instead of submitting PR right away since there are a few things to discuss: 

1) There are 2 ways to implement that:
    a) one is to create an additional type and a global user-settable switch that would instantiate 'pseudo' instead of 'half' - what would be a good name for the new type? 'pseudo-half'? 
    b) another way is to do the switch only and never expose additional type.

a) has an advantage of user to only select some subset of data to operate in pseudo mode - this have proven to be very useful in caffe experiments. for example, we needed to introduce 3 user switches in  caffe : to use 'pseudo' for forward and back filter/data convolutions. If we had an explicit pseudo in torch, it can be expressed explicitly in Lua code and no cudnn switches would be necessary.

2) Global 'pseudo' switch would operate the same way for both plans a) and b): all variables declared 'half' would in fact be instantiated as 'pseudo-half' and use float math and half storage. What woudl be a good name for that?  I am proposing 'cutorch.forceFloatMathForHalf' or 'cutorch.forcePseudoFP16'.

@soumith: let me know what you think!

borisfom:
Update: for the new type above I think the following will work quite well:
/* "pseudo-fp16" type: 16-bit storage, float math */
struct half_float {
  half val;
};

borisfom:
Copying here from CUNN discussion https://github.com/torch/cunn/issues/347
I was thinking, Torch could do better than global runtime flag. 
[...]

PR #530, opened 2016-10-07
Move more functions into generic

killeent:
This PR moves the following Tensor functions into the generic directory:

- mean/meanall are extended to all tensor types
- clamp to all tensor types
- cross to all tensor types
- cmin/cmax to all tensor types
- addcmul/addcdiv to all tensor types
- std/stdall to float/half/double tensors
- var/varall to float/half/double tensors
- norm/normall to float/half/double tensors
- renorm to float/half/double tensors
- lerp to float/half/double tensors
- Required refactoring a lot of float specific code into .cuh files.

Test Plan:

updated unit tests to compare using new tensor types


borisfom:
@killeent: can we coordinate that with https://github.com/torch/cutorch/issues/515 ?

borisfom:
Also, THCNumeric.h is getting way too much of code duplication. Let me submit a PR for refactored version that would be a small fraction of current size and allow for less changes going forward.

soumith:
i will be merging this, as it's a big functional PR, and it brings a lot of features in. Delaying this will really result in merge conflicts and really hard refactoring work.

working with #515 should follow.

borisfom:
@soumith: sure, merging this fast what I was hoping for! There is nothing contradictory between this and pseudo-float work, except that we touch the same files and moving stuff around.

PR #533, opened 2016-10-08
replace long with ptrdiff_t for memory size/offset, element count

BTNC:
Hi,

This is a equivalent PR with changes in torch, all tests passed. It is expected to introduce no behavior change for most *nix users.

Thanks,


soumith:
Thanks!

borisfom:
Shouldn't size and count be size_t ?

BTNC:
It should better be size_t, but the signed to unsigned change will introduce interface/behavior changes in many places which may surprise users. I have detailed why I choose to use ptrdiff_t in a equivalent [PR for torch](https://github.com/torch/torch7/pull/785). After read the torch PR, you may know how I decoupled the task of replacing long into several sub-tasks. The very first step is to identify which 'long' is about size, offset, count and replace with ptrdiff_t. Whether to replace (some) ptrdiff_t with size_t is up to the community but the final decision should be made by the admin. Most of the rest 'long' could be replaced with fixed width integers or a customizable torch defined 'LONG' type, and this is also up to the admin. Actually if the decision is made, I would be glad to take it over. Changes after this ptrdiff_t PR should be relatively smooth.

By the way, this PR's background is I am using msvc whose long is 32bit on x64 windows, and I want to enlarge my memory limit without surprise current users especially *nix users.

PR #543, opened 2016-10-13
changes to make cunn compile on windows with msvc

BTNC:
Hi,

This PR is to get cunn compile with msvc (at least with msvc19, the latest one).

Thanks,


soumith:
Thanks!

PR #546, opened 2016-10-13
Revert "changes to make cunn compile on windows with msvc"

soumith:
Reverts torch/cutorch#543

@BTNC I am reverting this, this breaks cutorch installation on Linux and OSX


BTNC:
Sorry I forgot to define THC_CLASS for non win32 case. I will fix it.

Issue #544, opened 2016-10-13
Math operations don't work with CudaHalfTensor on a Pascal card (1080)


ngimel:
```
th> a=torch.CudaHalfTensor(2,2):fill(2)
th> a:mul(2)
 0  0
 0  0
[torch.CudaHalfTensor of size 2x2]
th> a:fill(2)
 2  2
 2  2
[torch.CudaHalfTensor of size 2x2]
th> a:add(2)
 2  2
 2  2
[torch.CudaHalfTensor of size 2x2]
```
This is not caught by the tests, because tests only test cmul (rather than mul), and for add they test only adding tensors, not adding number to a tensor. It would be good to merge https://github.com/torch/cutorch/pull/521 or at least disable fp16 math for simple operations while leaving it on for hgemm. 
I have a fresh install from torch-distro (that points to 6-day old commit of cutorch, aa13020), luarocks install cutorch fails as in https://github.com/torch/cutorch/issues/541

soumith:
@ngimel fixed the master cutorch failing

soumith:
@ngimel i will fix this bug right now but i need some clarification on what you say about disabling fp16 math.

https://github.com/torch/cutorch/pull/521/files seems like more of a hack than anything. Our P100 users wont magically know that they can now enable this flag. 
Is there a way to find out whether the cards have half instructions at "slow" rate or at full rate? (if it's a small program that can check for this, i can add it to the cmake build)

ngimel:
P100 has compute capability 6.0, all other Pascals are 6.1, that can be used to differentiate between them. 
I'd say that this
https://github.com/torch/cutorch/blob/master/lib/THC/THCHalf.h#L13
has to be defined to 0 always (it is used in bandwidth-bound pointwise tensor operations, and wouldn't have any performance advantages even on P100. Besides, it is buggy, as I've shown in this issue, and it is still buggy with cutorch master). THC_nativeHalfInstructions (that's used to determine which gemm version to use) can return 1 on P100 and 0 on other Pascals.

soumith:
@ngimel ok i'm making these changes shortly. apart from the add bug, if you find anything else, do report, and we'll fix.
Thanks.

Issue #541, opened 2016-10-12
'luarocks install cutorch' error on Ubuntu 14.04


satwikkottur:
Gets stuck at 39% with several such messages:

[ 39%] /tmp/luarocks_cutorch-scm-1-9484/cutorch/lib/THC/THCTensorRandom.cu(250): error: more than one instance of overloaded function "min" matches the argument list:
            function "min(int, int)"
            function "min(unsigned int, unsigned int)"
            function "min(int, unsigned int)"
            function "min(unsigned int, int)"
            function "min(long long, long long)"
            function "min(unsigned long long, unsigned long long)"
            function "min(long long, unsigned long long)"
            function "min(unsigned long long, long long)"
            function "min(float, float)"
            function "min(double, double)"
            function "min(float, double)"
            function "min(double, float)"
            argument types are: (ptrdiff_t, ptrdiff_t)

soumith:
what CUDA version are you on?

satwikkottur:
CUDA v7.0

soumith:
i'm checking it now

soumith:
fixed now

BTNC:
I'm very sorry for the compilation error, it is an error introduced by my ptrdiff_t change. I will fix it.

ambarpal:
@BTNC @soumith 
I think the issue is not fixed completely. I'm still getting the error on Ubuntu 15.04, CUDA v7.0
I installed torch from scratch, following instructions on the getting started page

> /home/ambar/torch/extra/cutorch/lib/THC/THCTensorRandom.cu(250): error: more than one instance of overloaded function "min" matches the argument list:
>             function "min(int, int)"
>             function "min(unsigned int, unsigned int)"
>             function "min(int, unsigned int)"
>             function "min(unsigned int, int)"
>             function "min(long long, long long)"
>             function "min(unsigned long long, unsigned long long)"
>             function "min(long long, unsigned long long)"
>             function "min(unsigned long long, long long)"
>             function "min(float, float)"
>             function "min(double, double)"
>             function "min(float, double)"
>             function "min(double, float)"
>             argument types are: (ptrdiff_t, ptrdiff_t)
> .....
> /home/ambar/torch/extra/cutorch/lib/THC/THCTensorRandom.cu(323): error: more than one instance of overloaded function "min" matches the argument list:
>             function "min(int, int)"
>             function "min(unsigned int, unsigned int)"
>             function "min(int, unsigned int)"
[...]

PR #548, opened 2016-10-14
make cunn compile with msvc && fix compilation failure for linux/mac os

BTNC:
Hi,

I'm very sorry my last PR broke *nix users compilation. This PR includes last PR and fixes to those compilation failures. Since I am not on Linux/Max OS and I can not see full compilation output across those issue descriptions, I hope this PR could really fix those failures. Specifically, the fix includes:
(1) define THC_CLASS for non win32 case, which shoule be the cause of #545
(2) revert an int -> ptrdiff_t change which maybe the cause of #541, #544, #551

I will resend the cunn PR after this cutorch PR is no longer causing compilation issues for Linux/Max OS users. Please send me the error output if you still get compilation error.

Thanks,

### DIFF
--- a/lib/THC/THCDeviceTensor-inl.cuh
+++ b/lib/THC/THCDeviceTensor-inl.cuh
@@ -28,7 +28,11 @@ template <typename T, int Dim,
           typename IndexT, template <typename U> class PtrTraits>
 __host__ __device__
 THCDeviceTensor<T, Dim, IndexT, PtrTraits>::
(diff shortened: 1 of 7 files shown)


soumith:
Thanks @BTNC

BTNC:
I am sorry for the trouble brought by my recent PR, and hope this PR can end those compilation issues. Thanks very much for your wonderful works!

PR #549, opened 2016-10-14
Fix caching allocator when used from multiple Lua threads

### LINKED ISSUE #539: THCCachingAllocator does not always work with sharedserialize'd threads
A situation that arises, e.g.,  when using optnet + DataParallelTable + THC_CACHING_ALLOCATOR, minimum example with comments below:
```lua
      require 'cunn'
      require 'cutorch'
      require 'cudnn'
      local nGPU = 2
      --allocate something
      local a = torch.CudaTensor(1000,1000):fill(1)
      --create a simple model
      local model = nn.Sequential()
      local l1 = cudnn.SpatialConvolution(32,32,3,3)
      model:add(l1)
      model:cuda()
      local input = torch.CudaTensor(32,32,14,14)
      --prop model forward, as in optnet      
      local out = model:forward(input)
--now, wrap model in dpt   
      local gpus = torch.range(1, nGPU):totable()
      local fastest, benchmark = cudnn.fastest, cudnn.benchmark

      local dpt = nn.DataParallelTable(1, true, true)
         :add(model, gpus)
         :threads(function()
            local cudnn = require 'cudnn'
            cudnn.fastest, cudnn.benchmark = fastest, benchmark
         end)
--at this point, there are 3 THCCachingAllocators, and they do not share allocated_blocks      
--and there are 3 lua_States that share storages due to sharedserialize, 
--so storage from the main thread can be garbage-collected in the spawned thread
--which has its own THCCachingAllocator and won't find this storage in its allocated_blocks
--leading to cudaError. 
      dpt.gradInput = nil
      model = dpt:cuda()
      model:forward(input)
```
when run with THC_CACHING_ALLOCATOR set to 1, the output is
THCudaCheck FAIL file=/workspace/ALL/torch-distro/extra/cutorch/lib/THC/generic/THCStorage.c line=147 error=17 : invalid device pointer
W/o caching allocator, runs without problem.

colesbury:
Use a single, global THCCachingAllocator instance.

Previously, each Lua thread had its own THCCachingAllocator instance.
However, threads can share storages, which means a segment could be
allocated from on THCCachingAllocator and freed on another, which
breaks.

Fixes #539

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,6 +32,8 @@ Set the environment variable `THC_CACHING_ALLOCATOR=1` to enable the caching CUD
 
 By default, cutorch calls `cudaMalloc` and `cudaFree` when CUDA tensors are allocated and freed. This is expensive because `cudaFree` synchronizes the CPU with the GPU. Setting `THC_CACHING_ALLOCATOR=1` will cause cutorch to cache and re-use CUDA allocations to avoid synchronizations.
 
+With the caching memory allocator, allocations and frees should logically be considered "usages" of the memory segment associated with streams, just like kernel launches. The programmer must insert the proper synchronization if memory segments are used from multiple streams.
+
 ###`cutorch.*` API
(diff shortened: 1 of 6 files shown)


soumith:
Thanks!

PR #550, opened 2016-10-14
Add stream API that is not based on indices

colesbury:
This implements the THC code so that we can expose streams as objects
instead of simply referring to them by indices. This is not exposed in
Lua yet.


soumith:
are there incompatible API changes in this PR?
For example, will cunn break?: https://github.com/torch/cunn/search?utf8=%E2%9C%93&q=stream

soumith:
i checked, and it's fine.

PR #555, opened 2016-10-17
Fix cutorch.getStream()

colesbury:
state->numUserStreams does not include the NULL stream, which is stored
in res->streams[i]


soumith:
Thanks!

PR #558, opened 2016-10-19
Add generic type support for toDeviceTensor.

gchanan:
This is necessary to support generic modules in cunn, which currently use toDeviceTensor with only CudaTensors.


soumith:
Thanks!

PR #561, opened 2016-10-24
Store the device in THCStorage

colesbury:
THCStorage_(getDevice) currently calls cudaPointerGetAttributes() to determine the storage's device. This call is relatively slow (~1 microsecond). Instead, store the device in the THCStorage struct.


soumith:
See #566

PR #562, opened 2016-10-24
we can use _NPROCESSOR_CONF instead of a hack for tegra

howard0su:


howard0su:
I suggest we add -j$(ncpu) to luarocks so that every project can build such way

soumith:
tegra runs out of memory if we do _NPROCESSOR_CONF, that's why this change was introduced.

howard0su:
It works fine for me at least this morning. I worried if we original is working around NPROCESS_ONLN is 1 then we manually hack around it to set it to 3.

Building on 4 cores
-- Found Torch7 in /home/ubuntu/torch/install
-- MAGMA not found. Compiling without MAGMA support
-- Autodetected CUDA architecture(s): 3.2
-- got cuda version 6.5
..........................
[ 82%] [ 83%] [ 84%] [ 85%] Building C object lib/THC/CMakeFiles/THC.dir/THCAllocator.c.o
Building C object lib/THC/CMakeFiles/THC.dir/THCGeneral.c.o
Building CXX object lib/THC/CMakeFiles/THC.dir/THCCachingAllocator.cpp.o
Building C object lib/THC/CMakeFiles/THC.dir/THCStorage.c.o
[ 86%] Building C object lib/THC/CMakeFiles/THC.dir/THCStorageCopy.c.o
[ 88%] [ 89%] Building C object lib/THC/CMakeFiles/THC.dir/THCTensor.c.o
Building C object lib/THC/CMakeFiles/THC.dir/THCStream.c.o
[ 90%] Building C object lib/THC/CMakeFiles/THC.dir/THCTensorCopy.c.o
[ 91%] Building C object lib/THC/CMakeFiles/THC.dir/THCThreadLocal.c.o
Linking CXX shared library libTHC.so
[ 91%] Built target THC
[ 92%] Generating TensorMath.c
Scanning dependencies of target cutorch
[ 94%] [ 95%] Building C object CMakeFiles/cutorch.dir/Storage.c.o
[ 97%] [ 96%] Building C object CMakeFiles/cutorch.dir/init.c.o
Building C object CMakeFiles/cutorch.dir/Tensor.c.o
Building C object CMakeFiles/cutorch.dir/TensorMath.c.o
[ 98%] Building C object CMakeFiles/cutorch.dir/TensorOperator.c.o
[100%] Building C object CMakeFiles/cutorch.dir/torch/utils.c.o
Linking C shared module libcutorch.so
[100%] Built target cutorch

soumith:
From here: https://www.gnu.org/software/libc/manual/html_node/Processor-Resources.html
NPROCESSOR_CONF should return the same value as NPROCESSOR_ONLN on tegra

howard0su:
it is not true for the platform that supports dynamic power-off cores. Tegra is one. When the system is idle, NPROCESS_ONLN=1 while NPROCESSOR_CONF=4. When you are running some workload, the system will turn on more cores and you may get NPROCESS_ONLN=4 while NPROCESSOR_CONF is still 4.

The same change causes issue with OpenBLAS (https://github.com/xianyi/OpenBLAS/pull/981) and torch/distro to detect the right number of cores in such platform. this is relatively common on ARM platforms.

soumith:
thanks!

PR #569, opened 2016-10-26
Use void for no-arg functions

colesbury:


lib/THC/THCCachingAllocator.cpp
```diff
@@ -306,7 +306,7 @@ static THCDeviceAllocator device_allocator = {
   &caching_allocator
 };
 
-THC_API THCDeviceAllocator* THCCachingAllocator_get()
+THC_API THCDeviceAllocator* THCCachingAllocator_get(void)
```
apaszke: This strict prototype thing is a problem only for C, there's no need to add this to `.cpp` files. It doesn't hurt though, we can leave it.

PR #580, opened 2016-11-02
Add supporting code for CUDA IPC

colesbury:
This adds three small pieces to help with sharing THCStorages across
processes:

 1. THCIpcAllocator: a THCDeviceAllocator to close shared memory handles in the
    child process.
 2. THCCachingAllocator_getBaseAllocation which returns the pointer and
    size of the underlying cudaMalloc allocation. This is necessary
    because cudaIpcGetMemHandle requires 'base' pointers
 3. Support for TH_STORAGE_VIEW in THCStorage_(free). This is useful in
    child processes to represent THCCachingAllocator allocations split
    from a larger cudaMalloc call.

### DIFF
--- a/lib/THC/THC.h
+++ b/lib/THC/THC.h
@@ -4,6 +4,7 @@
 #include "THCGeneral.h"
 #include "THCAllocator.h"
 #include "THCBlas.h"
+#include "THCCachingAllocator.h"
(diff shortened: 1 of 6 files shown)


soumith:
Thanks!

PR #582, opened 2016-11-03
THC UVA Allocator

nicolasvasilache:
This PR adds support for UVA allocation (backed by virtual memory). 

Depending on compute capabilities, perf and coherence vary so I am just adding the possibility to use UVA. 

Since a single allocation results in a pointer that can be used directly both on CPU and GPU, the conversion function only reference the same pointer and never allocate.

Similarly to the current cudaHostAllocator, only Cuda/Float tensors are supported atm.

I have not propagated the pointer passing to the tensor:cuda() and tensor:float() functions, this will be for a future diff.

### DIFF
--- a/FFI.lua
+++ b/FFI.lua
@@ -32,6 +32,7 @@ typedef struct THCState
   int numUserStreams;
   int numUserBlasHandles;
   struct THAllocator* cudaHostAllocator;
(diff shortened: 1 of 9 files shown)


soumith:
you'll have to rebase, as well as remove debug prints

nicolasvasilache:
yes sorry, will redo

PR #583, opened 2016-11-03
THC UVA Allocator

nicolasvasilache:
This PR adds support for UVA allocation (backed by virtual memory).

Depending on compute capabilities, perf and coherence vary so I am just adding the possibility to use UVA.

Since a single allocation results in a pointer that can be used directly both on CPU and GPU, the conversion function only reference the same pointer and never allocate.

Similarly to the current cudaHostAllocator, only Cuda/Float tensors are supported atm.

I have not propagated the pointer passing to the tensor:cuda() and tensor:float() functions, future work.

### DIFF
--- a/init.c
+++ b/init.c
@@ -891,6 +891,24 @@ static int cutorch_setHeapTracking(lua_State *L)
   return 0;
 }
 
+static int cutorch_isManagedPtr(lua_State *L)
+{
+  THCState *state = cutorch_getstate(L);
+  if(lua_type(L, 1) != LUA_TNUMBER) {
+    THError("Must receive a ptr cast as a number");
+  }
+  void* ptr = (void* )luaL_optinteger(L, 1, 0);
+  struct cudaPointerAttributes attributes;
+  cudaError_t res = cudaPointerGetAttributes(&attributes, ptr);
+  if (res == cudaErrorInvalidValue) {
+    lua_pushboolean(L, 0);
+  } else {
+    THCudaCheck(res);
+    lua_pushboolean(L, attributes.isManaged);
+  }
+  return 1;
+}
+
 static int cutorch_shutdown(lua_State *L)
 {
   THCState **state = (THCState **) lua_topointer(L, 1);
@@ -911,11 +929,11 @@ static int cutorch_hasHalfInstructions(lua_State *L) {
 
 static int cutorch_hasFastHalfInstructions(lua_State *L) {
   THCState *state = cutorch_getstate(L);
-#ifdef CUDA_HALF_TENSOR  
+#ifdef CUDA_HALF_TENSOR
   lua_pushboolean(L, THC_fastHalfInstructions(state));
 #else
   lua_pushboolean(L, 0);
-#endif  
+#endif
   return 1;
 }
 
@@ -957,6 +975,7 @@ static const struct luaL_Reg cutorch_stuff__ [] = {
   {"setRNGState", cutorch_setRNGState},
   {"getState", cutorch_getState},
   {"setHeapTracking", cutorch_setHeapTracking},
+  {"isManagedPtr", cutorch_isManagedPtr},
   {NULL, NULL}
 };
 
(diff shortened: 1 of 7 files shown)


init.c
```diff
@@ -891,6 +891,19 @@ static int cutorch_setHeapTracking(lua_State *L)
...
 
+static int cutorch_isManagedPtr(lua_State *L)
+{
+  THCState *state = cutorch_getstate(L);
+  if(lua_type(L, 1) != LUA_TNUMBER) {
+    THError("Must receive a ptr cast as a number");
+  }
+  void* ptr = (void* )luaL_optinteger(L, 1, 0);
+  struct cudaPointerAttributes attributes;
+  cudaPointerGetAttributes (&attributes, ptr);
```
colesbury: cudaPointerGetAttributes returns a `cudaError_t`, which should be checked. For CPU memory that's not managed or pinned it returns "cudaErrorInvalidValue", which should probably just be translated to `false` at the user level.

init.lua
```diff
@@ -48,11 +46,81 @@ function cutorch.createCudaHostTensor(...)
...
+-- Accepts either a LongStorage or a sequence of numbers.
+function cutorch.createCudaHostTensor(...)
+   local size = longTensorSize(...)
    local storage = torch.FloatStorage(cutorch.CudaHostAllocator, size:prod())
    return torch.FloatTensor(storage, 1, size:storage())
 end
 
+-- Creates a CudaTensor using the CudaUVAAllocator.
+-- Accepts either a LongStorage or a sequence of numbers.
+function _createUVATensor(...)
```
colesbury: Can this be `local`?

init.lua
```diff
@@ -48,11 +46,81 @@ function cutorch.createCudaHostTensor(...)
...
+      t:storage():size(),
+      tonumber(torch.data(t:storage(), true))
+   )
+   cutorch.streamSynchronize(0)
+   assert(cutorch.isManaged(storage))
+   return torch.CudaTensor(storage)
+end
+
+function cutorch.isManaged(t)
+   if not torch.type(t):find('Tensor') and not torch.type(t):find('Storage') then
```
colesbury: I think `torch.isTensor` and `torch.isStorage` should work

init.lua
```diff
@@ -48,11 +46,81 @@ function cutorch.createCudaHostTensor(...)
...
+      tonumber(torch.data(ft:storage(), true))
+   )
+   return torch.CudaTensor(storage)
+end
+
+-- UVA storage is a single memory location backed by virtual addressing.
+-- Converting between CPU / GPU tensor types is done by raw pointer passing.
+-- We only support FloatTensor, CudaTensor, Cuda -> float and float -> Cuda atm
+function cutorch.toFloatUVATensor(t)
+   if not torch.type(t):find('Tensor') then
```
colesbury: `torch.isTensor`

init.lua
```diff
@@ -48,11 +46,81 @@ function cutorch.createCudaHostTensor(...)
...
+      error('Must use a tensor, got ' .. torch.type(t))
+   end
+   -- See THCAllocator.c for explanation about synchronization
+   cutorch.streamSynchronize(0)
```
colesbury: Why the `cutorch.streamSynchronize(0)`? Also I don't see any explanation in THCAllocator.c about synchronization

nicolasvasilache:
Cool thanks I'll make the changes and reference back to CUDA_C_Programming_Guide.pdf section J.

In short, pre-Pascal requires you to synchronize or you will see crashes, for more details see the pdf

"It is worth a comment on the synchronization between host and device. Notice how in
the non-managed example, the synchronous cudaMemcpy() routine is used both to
synchronize the kernel (that is, to wait for it to finish running), and to transfer the data
to the host. The Unified Memory examples do not call cudaMemcpy() and so require an
explicit cudaDeviceSynchronize() before the host program can safely use the output
from the GPU."

colesbury:
OK, so if you modify managed memory on the GPU, you have to call synchronize before you access the managed memory on the CPU. But why do you have device synchronization calls around the allocation? These do not exist in the CUDA managed memory examples.

nicolasvasilache:
Because in the examples in the pdf, cudaMallocManaged is the first thing happening in the main?

Section J.2.2.1 says
"
Note that if memory is dynamically allocated with cudaMallocManaged() or
cuMemAllocManaged() while the GPU is active, the behavior of the memory is
unspecified until additional work is launched or the GPU is synchronized. Attempting
to access the memory on the CPU during this time may or may not cause a segmentation
fault."

PR #589, opened 2016-11-09
Move random functions to generic

killeent:
This diff moves random-number related sampling functions to generic, including:

- bernoulli
- cauchy
- geometric
- multinomial
- normal
- rand
- uniform
- exponential
- logNormal
- randn

Test Plan:
- refactored unit tests to work with multiple cuda tensor types

Some Notes for review:

1) For the new templatized multinomial code, I had to go from:

`extern __shared__ float smem[];`

to:

`extern __shared__ __align__(sizeof(T)) unsigned char my_smem[];`
`T *smem = reinterpret_cast<T *>(my_smem);`

I'm not sure if this is the best way of doing things, I got it from StackOverflow.

2) I used the _double curand methods where appropriate.
3) Some tests are more flakey after this diff - for example, the tests that sample multinomial with replacement sometimes fail for CudaHalfTensor.
4) We now have parity with the torch implementation of logNormal. Previously, this was broken - see discussion in: https://github.com/torch/torch7/pull/834#issuecomment-259991279


test/test.lua
```diff
@@ -2567,10 +2578,13 @@ function test.logNormal()
    local tolerance = 0.01
    local t = torch.CudaTensor(sz1, sz2)
 
-   t:logNormal(mean, std)
-   local logt = t:log()
-   tester:assertalmosteq(logt:mean(), mean, tolerance, "mean is wrong")
-   tester:assertalmosteq(logt:std(), std, tolerance, "standard deviation is wrong")
+   for _, typename in ipairs(float_typenames) do
```
soumith: test:
x:mean() == mean and x:std() == std
logx = torch.log(x)
logx:mean() == mean * mean / math.sqrt(std * std + mean * mean)
logx:std() == math.sqrt(math.log(std * std / mean * mean + 1)

PR #592, opened 2016-11-10
fix: cunn can't find cutorch sources

joker512:
https://github.com/torch/distro/issues/138#issuecomment-259133935


soumith:
@joker512 i'm not sure why you want this merged. this will prob work in your particular setting, but will likely break torch distro installs

joker512:
Ok, I'll explain it step by step:

1. https://github.com/torch/cutorch/blob/master/lib/THC/CMakeLists.txt :97-114 and _luarocks install cutorch_ install cutorch sources to **/usr/include/THC**

1. https://github.com/torch/cunn/blob/master/CMakeLists.txt :9 and _luarocks install cunn_ force cunn to search cutorch sources in **/usr/include/lua5.x/THC**

1. It's incompatible. Source for cutorch can't be found by cunn:
```
[  1%] Building NVCC (Device) object lib/THCUNN/CMakeFiles/THCUNN.dir/THCUNN_generated_VolumetricReplicationPadding.cu.o
/home/strannik/soft/nn/cunn/lib/THCUNN/VolumetricReplicationPadding.cu:3:31: fatal error: THCDeviceTensor.cuh: No such file or directory
compilation terminated
```
Both by _make_ and _luarocks install_.

I guess that since everybody installs torch with **distro** package problem didn't emerge before. In any case the code:
```
IF(NOT THC_INSTALL_BIN_SUBDIR
    OR NOT THC_INSTALL_LIB_SUBDIR
    OR NOT THC_INSTALL_INCLUDE_SUBDIR
    OR NOT THC_INSTALL_CMAKE_SUBDIR)
...
ELSE()
  SET(THC_INSTALL_BIN_SUBDIR ${Torch_INSTALL_BIN_SUBDIR})
  SET(THC_INSTALL_LIB_SUBDIR ${Torch_INSTALL_LIB_SUBDIR})
  SET(THC_INSTALL_INCLUDE_SUBDIR ${Torch_INSTALL_INCLUDE_SUBDIR})
  SET(THC_INSTALL_CMAKE_SUBDIR ${Torch_INSTALL_CMAKE_SUBDIR})
ENDIF()
```
Looks strange. Why do we need to set THC_INSTALL_* variables just to reset them to **/usr/include/lua5.x/THC** dir?
For torch package everything is correct: https://github.com/torch/torch7/blob/master/lib/CMakeLists.txt
```
SET(TH_INSTALL_BIN_SUBDIR "${Torch_INSTALL_BIN_SUBDIR}")
SET(TH_INSTALL_LIB_SUBDIR "${Torch_INSTALL_LIB_SUBDIR}")
SET(TH_INSTALL_INCLUDE_SUBDIR "${Torch_INSTALL_INCLUDE_SUBDIR}")
SET(TH_INSTALL_CMAKE_SUBDIR "${Torch_INSTALL_CMAKE_SUBDIR}")

ADD_SUBDIRECTORY(TH)
ADD_SUBDIRECTORY(luaT)
```
The package installation path is **/usr/include/lua5.x/TH** without any conditions. Why cutorch package is so different?

I've spent a lot of hours to sort it out. And I'm sure it is a bug that prevents installing cunn to the system by luarocks if I don't want to install it by distro metapackage. Please, explain me if I am not correct.

soumith:
thanks for the PR.
I had to spend quite some time understanding the problem as well (before merging), so hence the delay.

PR #594, opened 2016-11-11
change to compile with msvc after THCState refactor

BTNC:


BTNC:
need more change to get cunn compile

PR #598, opened 2016-11-14
move random functions to generic (attempt 2)

killeent:
Same as https://github.com/torch/cutorch/pull/589 but with fixes to:

- bernoulli test code
- multinomial test code
- geometric implementation and test


soumith:
Thanks!

PR #600, opened 2016-11-15
Implemented cudaMemGetInfo for caching allocator

borisfom:
I tested it  with test_shutdown.lua


borisfom:
@soumith, @szagoruyko : this fix is ready to go in.

lib/THC/THCCachingAllocator.cpp
```diff
@@ -158,12 +158,12 @@ struct THCCachingAllocator
     allocated_blocks.erase(it);
```
soumith: i think @colesbury had some reservations about this as well. this can be greatly simplified. i'll let him comment on monday.

init.c
```diff
@@ -694,27 +694,38 @@ static int cutorch_setKernelPeerToPeerAccess(lua_State *L)
...
+  size_t freeBytes = 0;
+
+  THCState *state = cutorch_getstate(L);
 
   int device = luaL_optint(L, 1, -10);
-  if (device == -10) { /* no argument passed, current device mem usage */
-    THCudaCheck(cudaMemGetInfo(&freeBytes, &totalBytes));
-  } else { /* argument was given, particular device's memory usage */
-    THCudaCheck(cudaSetDevice(device-1)); /* zero indexed */
-    THCudaCheck(cudaMemGetInfo(&freeBytes, &totalBytes));
```
borisfom: Right, me trying to avoid code duplication here led to code bloat instead :). Cleaned up.

lib/THC/THCCachingAllocator.cpp
borisfom: The only way to simplify it would be to introduce counter of free memory and update it on each allocation/deallocation.
That would be a sensible approach as well. I was shooting for a least intrusive method that does not add any complexity (and locks) to the main functionality.

borisfom: And , yes, the largest block size parameter is now unused.
@soumit, @colesbury: we can either purge it or hook it up for the user to actually get this info, let me know which way you prefer.

soumith:
Thanks!

PR #601, opened 2016-11-15
[cutorch] remove syncing point from baddbmm

1nadequacy:
This change removes HtoD copies inside baddbmm. These copies
introduce a syncing point which causes slow downs in a multi
gpu training.

Test plan: Run unittests for baddbmm.


soumith:
thanks!

PR #611, opened 2016-11-24
Revert "Lazily initialize CUDA devices"

soumith:
Reverts torch/cutorch#610

I'm backing out this change for now.
Tests for nn.GPU in cunn are failing with this.


soumith:
cc: @colesbury

colesbury:
So the bug was that I forget to check/enable p2p access in a few places where we might use p2p kernels. I'll add checks in those places, but it seems like we're often relying on p2p kernels in places where we might not expect it. I've document those in #612

Issue #612, opened 2016-11-25
Tensor / Storage copy bugs


colesbury:
We have a number of tensor/storage copy bugs, which we should fix:

1. Copies of non-contiguous tensors from GPU to host create a contiguous copy on the **current** device. It should create the contiguous copy on the same device as the original tensor. Otherwise, this can entail a p2p copy (or worse, a host mediated copy).

2. The above issue exists with CUDA half -> CPU float copies (even contiguous ones)

3. Cross-device storage copies do not insert the correct synchronizations (unlike cross-device tensor copies).

4. Float <-> Half storage copies between devices require P2P access, but does not check for it (as opposed to cudaMemcpy which falls back to host mediated copies)

Our copy code is duplicated and spread over a number of places. Storage copies are implemented separately from Tensor copies, even though storage copies are a (simple) special case of tensor copies. We should combine the implementations to reduce duplicate code and add tests for all type conversions, including cross-GPU type conversions.

borisfom:
We'd have much less of those copies if my HalfTensor PRs: https://github.com/torch/torch7/pull/826 and https://github.com/torch/cutorch/pull/578 were merged, too!

soumith:
these bugs are solved now, thanks to @colesbury 's PRs.

PR #613, opened 2016-11-25
Lazily initialize CUDA devices (take 2)

colesbury:
Previously, cutorch would initialize every CUDA device and enable P2P
access between all pairs. This slows down start-up, especially with 8
devices. Now, THCudaInit does not initialize any devices and P2P access
is enabled lazily. Setting the random number generator seed also does
not initialize the device until random numbers are actually used.

I've updated the Storage copy code to delegate the Tensor copy code. This 
fixes the issues with p2p not being enabled and adds proper inter-GPU
synchronization (see #612)

### DIFF
--- a/init.c
+++ b/init.c
@@ -776,35 +776,35 @@ static int cutorch_getDeviceProperties(lua_State *L)
 
 static int cutorch_seed(lua_State *L)
 {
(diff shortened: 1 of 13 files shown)


soumith:
thanks!

PR #614, opened 2016-11-28
use local modified select_compute_arch.cmake for msvc

BTNC:
Hi,

select_compute_arch.cmake is introduced into cmake 3.7, but it does not work with msvc. This change it to let cutorch stick with local select_compute_arch.cmake for msvc.

Thanks,


soumith:
thanks!

PR #617, opened 2016-11-30
Reducing generic type set via controlling macros, 

borisfom:
…  on by default on Android.
@soumith: this is what we discussed earlier. I have deliberately left copy/init part unconditional, so that only math operators are disabled. That has benefits: CUDNN Double test passes even with CudaDoubleTensor operations off.
With that change, THNN size also goes down ~2.5x on Android.


lib/THC/generic/THCTensorCopy.cu
```diff
@@ -31,16 +31,27 @@ THCTensor_(copyIgnoringOverlaps)(THCState* state, THCTensor* dst, THCTensor* src
   }
 
 IMPLEMENT_THC_CUDA_TENSOR_COPY(Byte, Byte)
+
+#ifndef TH_GENERIC_NO_CHAR
```
borisfom: Thanks for catching this: yes it seems I was compiling w/o those macros defined - apologize for the confusion.

CMakeLists.txt
```diff
@@ -12,6 +12,7 @@ SET(CMAKE_C_FLAGS "-std=c99 -Werror=implicit-function-declaration ${CMAKE_C_FLAG
 ENDIF (NOT WIN32)
 IF(CUDA_HAS_FP16 OR NOT ${CUDA_VERSION} LESS 7.5)
   SET(CMAKE_C_FLAGS "-DCUDA_HAS_FP16=1 ${CMAKE_C_FLAGS}")
+  SET(CMAKE_C_FLAGS "-DREDUCED_TYPE_SET=1 ${CMAKE_C_FLAGS}")
```
borisfom: Thanks for catching this: yes it was supposed to be THC_REDUCED_TYPE_SET, for testing purposes only.

borisfom:
Thanks for the review! I apologize for the confusion - looks like I did not compile correctly due to CMakelists issue. I have made this branch build now, but I agree the changes are too invasive and I will rethink the approach.

CMakeLists.txt
borisfom: Responding to questions:
1. "What's special about char, short, int, double to be considered for non-compilation, compared to e.g. byte?"
The goal was to provide a smaller size libraries for Android apps that use float tensors only.
 Code bloat from nvcc is huge. I have tried 'byte' and 'long' initially, too, but they were used internally by cutorch helper routines.

2. "The need to have #ifdef CUDA_HALF_TENSOR and #ifndef CUDA_NO_HALF for half types to be defined is going to have constant maintenance problems -- we already break cutorch regularly because of lack of CUDA_HALF_TENSOR..."
Actually,  hopefully CUDA_HALF_TENSOR could be retired soon - probably as soon as we have CPU Half integrated, so I don't see much issue here.
3. "Similar to 1) above, I don't see how people are going to remember that half (sort of), char, short, int, double are special and need to be guarded and other types don't."
I agree, those changes are not pretty, and require some re-thinking and/or refactoring beforehand. Part of the reason the changes are so uneven and scattered is because the code may be too expanded and some clever use of macros may help. 
I am going to withdraw this PR for now.

PR #621, opened 2016-12-02
Assure the refcount is greater than zero before decrementing

drahnr:
Otherwise one might do stupid things if using cuda backend in embedded lua and tensors both from lua and C


apaszke:
See my comments in torch/torch7#860

PR #633, opened 2016-12-14
	fix wrong export directive for THCCachingHostAllocator

elikosan:
It did prevent linking of THC.dll on Windows.


soumith:
thanks!

Issue #615, opened 2016-11-29
torch.cat needs to be optimized


soumith:
catArray needs to be optimized to be on-device.
Otherwise,  if we torch.cat 100s of small  tensors at the same time, the calls are all serialized and also have kernel launch overhead.

@killeent is working on this.

pavanky:
@soumith Is this going to be for `CudaTensor` only or for other Cuda Tensors as well ?

soumith:
The general improvement will be for all types

pavanky:
@soumith Thanks for the clarification. I did not check the progress made here: https://github.com/torch/cutorch/issues/529 :)

PR #635, opened 2016-12-15
First stab at parallelizing catArray to multiple tensors per kernel

killeent:
Addresses #615.

This diff implements a new kernel for catArray, that supports copying multiple input tensors into the output tensor at once. This is still a WIP, and a subsequent commit will add optimizations and tweaks after evaluating performance, along with some calculations. Just wanted to get this up so that @soumith and @wickedfoo could start taking a look if they wanted.

The basic idea for the kernel is to mimic the grid-stride loop and indexing logic from pointwiseApply, but to add a y component to the grid to represent which of the Tensors passed to kernel to operate on. We define a new struct CatArrayKernelParam, which allows passing multiple tensors to a single kernel.

Further optimizations and work I will explore:

- What we set gridDim.x to be
- Increasing CAT_ARRAY_KERNEL_BATCH_SIZE
- Collapsing Dims on TensorInfos
- Try to actually use smarter indexing as opposed to falling back to dynamic dims in IndexToOffset

Inevitable Downsides:

- The existing tensor to tensor copy code has a number of optimizations, including support for memcpy, when applicable. This is not maintained.

### DIFF
--- a/lib/THC/THCTensorInfo.cuh
+++ b/lib/THC/THCTensorInfo.cuh
@@ -18,10 +18,15 @@
 // CUDA kernel argument that defines tensor layout
 template <typename T, typename IndexType>
 struct TensorInfo {
-  TensorInfo(T* p,
+  __host__ __device__ TensorInfo(T* p,
              int dim,
              IndexType sz[MAX_CUTORCH_DIMS],
              IndexType st[MAX_CUTORCH_DIMS]);
+
+  // Default constructor so that we can use TensorInfos as fields in other structs.
+  // In general, TensorInfo's should be created via getTensorInfo(...) and not by
+  // manually setting the fields
+  TensorInfo();
 
   // Set the size of the given dimension to 1, as if it were a
   // reduction dim (allows you to calculate offsets of the reduction
@@ -65,6 +70,9 @@ TensorInfo<T, IndexType>::TensorInfo(T* p,
     strides[i] = st[i];
   }
 }
+
+template <typename T, typename IndexType>
+TensorInfo<T, IndexType>::TensorInfo() {}
 
 template <typename T, typename IndexType>
 void
@@ -237,7 +245,7 @@ TensorInfo<T, IndexType>::newNarrow(int dimension, IndexType firstIndex, IndexTy
   }
 
   T* newData = data + (firstIndex * strides[dimension]);
-  st[dimension] = size;
+  sz[dimension] = size;
 
   return TensorInfo<T, IndexType>(newData, dims, sz, st);
 }
--- a/lib/THC/THCTensorInfo.cuh
+++ b/lib/THC/THCTensorInfo.cuh
@@ -43,6 +43,8 @@ struct TensorInfo {
     return (dims == 1 && strides[0] == 1);
   }
 
+  __host__ __device__ TensorInfo<T, IndexType> newNarrow(int dimension, IndexType firstIndex, IndexType size) const;
+
   T* data;
   IndexType sizes[MAX_CUTORCH_DIMS];
   IndexType strides[MAX_CUTORCH_DIMS];
@@ -221,6 +223,23 @@ TensorInfo<T, IndexType>::collapseDims(int excludeDim) {
   // renumbered to this new `returnDim`, since some dimensions could
   // have been collapsed.
   return returnDim;
+}
+
+template <typename T, typename IndexType>
+TensorInfo<T, IndexType>
+TensorInfo<T, IndexType>::newNarrow(int dimension, IndexType firstIndex, IndexType size) const {
+  IndexType sz[MAX_CUTORCH_DIMS];
+  IndexType st[MAX_CUTORCH_DIMS];
+
+  for (int i = 0; i < dims; ++i) {
+    sz[i] = sizes[i];
+    st[i] = strides[i];
+  }
+
+  T* newData = data + (firstIndex * strides[dimension]);
+  st[dimension] = size;
+
+  return TensorInfo<T, IndexType>(newData, dims, sz, st);
 }
 
 // Translate a linear index for the apply to a T* offset;
--- a/lib/THC/THCTensorInfo.cuh
+++ b/lib/THC/THCTensorInfo.cuh
@@ -18,15 +18,10 @@
 // CUDA kernel argument that defines tensor layout
 template <typename T, typename IndexType>
 struct TensorInfo {
-  __host__ __device__ TensorInfo(T* p,
+  TensorInfo(T* p,
              int dim,
              IndexType sz[MAX_CUTORCH_DIMS],
              IndexType st[MAX_CUTORCH_DIMS]);
-
-  // Default constructor so that we can use TensorInfos as fields in other structs.
-  // In general, TensorInfo's should be created via getTensorInfo(...) and not by
-  // manually setting the fields
-  TensorInfo();
 
   // Set the size of the given dimension to 1, as if it were a
   // reduction dim (allows you to calculate offsets of the reduction
@@ -48,15 +43,6 @@ struct TensorInfo {
     return (dims == 1 && strides[0] == 1);
   }
 
-  // Returns a TensorInfo that is narrowed along the specified dimension,
-  // at the specified offset/size. Equivalent to narrow in THCTensor, in
-  // terms of its effect on the data pointer, and size/stride values.
-  // newNarrow is a little bit of a misnomer, as we are not actually
-  // allocating memory, but returning a 'new' TensorInfo that is the input
-  // TensorInfo narrowed.
-  __host__ __device__ TensorInfo<T, IndexType> newNarrow(
-      int dimension, IndexType firstIndex, IndexType size) const;
-
   T* data;
   IndexType sizes[MAX_CUTORCH_DIMS];
   IndexType strides[MAX_CUTORCH_DIMS];
@@ -77,9 +63,6 @@ TensorInfo<T, IndexType>::TensorInfo(T* p,
     strides[i] = st[i];
   }
 }
-
-template <typename T, typename IndexType>
-TensorInfo<T, IndexType>::TensorInfo() {}
 
 template <typename T, typename IndexType>
 void
@@ -238,24 +221,6 @@ TensorInfo<T, IndexType>::collapseDims(int excludeDim) {
   // renumbered to this new `returnDim`, since some dimensions could
   // have been collapsed.
   return returnDim;
-}
-
-template <typename T, typename IndexType>
-__host__ __device__
-TensorInfo<T, IndexType>
-TensorInfo<T, IndexType>::newNarrow(int dimension, IndexType firstIndex, IndexType size) const {
-  IndexType sz[MAX_CUTORCH_DIMS];
-  IndexType st[MAX_CUTORCH_DIMS];
-
-  for (int i = 0; i < dims; ++i) {
-    sz[i] = sizes[i];
-    st[i] = strides[i];
-  }
-
-  T* newData = data + (firstIndex * strides[dimension]);
-  sz[dimension] = size;
-
-  return TensorInfo<T, IndexType>(newData, dims, sz, st);
 }
 
 // Translate a linear index for the apply to a T* offset;
--- a/lib/THC/THCTensorInfo.cuh
+++ b/lib/THC/THCTensorInfo.cuh
@@ -48,7 +48,14 @@ struct TensorInfo {
     return (dims == 1 && strides[0] == 1);
   }
 
-  __host__ __device__ TensorInfo<T, IndexType> newNarrow(int dimension, IndexType firstIndex, IndexType size) const;
+  // Returns a TensorInfo that is narrowed along the specified dimension,
+  // at the specified offset/size. Equivalent to narrow in THCTensor, in
+  // terms of its effect on the data pointer, and size/stride values.
+  // newNarrow is a little bit of a misnomer, as we are not actually
+  // allocating memory, but returning a 'new' TensorInfo that is the input
+  // TensorInfo narrowed.
+  __host__ __device__ TensorInfo<T, IndexType> newNarrow(
+      int dimension, IndexType firstIndex, IndexType size) const;
 
   T* data;
   IndexType sizes[MAX_CUTORCH_DIMS];
@@ -234,6 +241,7 @@ TensorInfo<T, IndexType>::collapseDims(int excludeDim) {
 }
 
 template <typename T, typename IndexType>
+__host__ __device__
 TensorInfo<T, IndexType>
 TensorInfo<T, IndexType>::newNarrow(int dimension, IndexType firstIndex, IndexType size) const {
   IndexType sz[MAX_CUTORCH_DIMS];
--- a/lib/THC/THCTensorMath.cuh
+++ b/lib/THC/THCTensorMath.cuh
@@ -23,14 +23,28 @@ __global__ void THCTensor_copyToDiagonal(T* a, T* b, ptrdiff_t start, ptrdiff_t 
   }
 }
 
-#define CAT_ARRAY_KERNEL_BATCH_SIZE 2
+// Limited by the maximum size of kernel arguments (4kb)
+#define CAT_ARRAY_KERNEL_BATCH_SIZE 16
 
 template <typename T>
 struct CatArrayKernelParam {
+  // Tensors to copy into the output parameter
   TensorInfo<T, unsigned int> inputs[CAT_ARRAY_KERNEL_BATCH_SIZE];
+
+  // The offsets along the dimension in the output tensor where we
+  // should begin the copy, for each tensor
   int offsets[CAT_ARRAY_KERNEL_BATCH_SIZE];
+
+  // The of the dimension for each input tensor, used to specify
+  // how to narrow the output tensor
   int dimSizes[CAT_ARRAY_KERNEL_BATCH_SIZE];
+
+  // Number of elements in each tensor, for the grid-stride loop
+  // bound
   int nElements[CAT_ARRAY_KERNEL_BATCH_SIZE];
+
+  // Actual number of tensors in this param (may be less than the
+  // batch size)
   int count;
 };
 
--- a/lib/THC/THCTensorMath.cuh
+++ b/lib/THC/THCTensorMath.cuh
@@ -23,4 +23,31 @@ __global__ void THCTensor_copyToDiagonal(T* a, T* b, ptrdiff_t start, ptrdiff_t 
   }
 }
 
+#define CAT_ARRAY_KERNEL_BATCH_SIZE 2
+
+template <typename T>
+struct CatArrayKernelParam {
+  TensorInfo<T, unsigned int> inputs[CAT_ARRAY_KERNEL_BATCH_SIZE];
+  int offsets[CAT_ARRAY_KERNEL_BATCH_SIZE];
+  int dimSizes[CAT_ARRAY_KERNEL_BATCH_SIZE];
+  int nElements[CAT_ARRAY_KERNEL_BATCH_SIZE];
+  int count;
+};
+
+template <typename T>
+__global__ void catArrayBatchedCopy(TensorInfo<T, unsigned int> result, const CatArrayKernelParam<T> param, int dimension) {
+  // A block is responsible for the ith tensor in the param if its blockDim.y = i, so let's narrow
+  // the result TensorInfo according to the offset, dimSize for that tensor
+  TensorInfo<T, unsigned int> nt = result.newNarrow(dimension, param.offsets[blockIdx.y], param.dimSizes[blockIdx.y]);
+
+  // Now follow the normal pointwise op code, where the the linear index is determined by thread/block x values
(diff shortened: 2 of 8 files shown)


lib/THC/THCTensorMath.cuh
```diff
@@ -23,4 +23,31 @@ __global__ void THCTensor_copyToDiagonal(T* a, T* b, ptrdiff_t start, ptrdiff_t
...
+  // the result TensorInfo according to the offset, dimSize for that tensor
+  TensorInfo<T, unsigned int> nt = result.newNarrow(dimension, param.offsets[blockIdx.y], param.dimSizes[blockIdx.y]);
+
+  // Now follow the normal pointwise op code, where the the linear index is determined by thread/block x values
+  for (unsigned int linearIndex = blockIdx.x * blockDim.x + threadIdx.x;
+       linearIndex < param.nElements[blockIdx.y];
+       linearIndex += gridDim.x * blockDim.x) {
+    const unsigned int resultOffset = IndexToOffset<T, unsigned int, -1>::get(linearIndex, nt);
+    const unsigned int srcOffset = IndexToOffset<T, unsigned int, -1>::get(linearIndex, param.inputs[blockIdx.y]);
+    nt.data[resultOffset] = param.inputs[blockIdx.y].data[srcOffset];
```
pavanky: I don't think it'll affect the performance much, but it'd be nice if we can store `param.inputs[blockIdx.y]` in a register outside the for loop.

lib/THC/THCTensorMath.cuh
```diff
@@ -23,4 +23,31 @@ __global__ void THCTensor_copyToDiagonal(T* a, T* b, ptrdiff_t start, ptrdiff_t
   }
 }
 
+#define CAT_ARRAY_KERNEL_BATCH_SIZE 2
```
pavanky: This should not be hardcoded.

lib/THC/generic/THCTensorMath.cu
```diff
@@ -122,17 +122,71 @@ void THCTensor_(catArray)(THCState *state, THCTensor *result,
...
+        param.offsets[j] = index;
+        param.dimSizes[j] = dimension < THCTensor_(nDimension)(state, inputs[i+j])
+               ? THCTensor_(size)(state, inputs[i+j], dimension)
+               : 1;
+        index += param.dimSizes[j];
+        param.nElements[j] = THCTensor_(nElement)(state, inputs[i+j]);
+        cohortMax = cohortMax > param.nElements[j] ? cohortMax : param.nElements[j];
+      }
+      param.count = j;
+
```
pavanky: Consider using constant memory to store `param` so it more arrays can be copied in a single kernel launch?

lib/THC/THCTensorMath.cuh
pavanky: 1) You can use the constant memory to store the pointers to all the arrays at the same time (assuming there are not tens of thousands of arrays you are cat'ing at the same time).

2) Pass in an array of `long`s which contain the values of the pointers to the arrays. The number of arrays can be equal to the number of blocks along y.

pavanky: Those can be in cosntant memory / global memory too.

killeent: So I am not super familiar with constant memory, but it seems to me like you do something like:

`__constant__ int deviceMem[128];`

and populate it via a call to `cudaMemcpyToSymbol(...)`. However, given that our TensorInfo structs have a template parameter, I am not sure if we can use this interface, as the size of the array would not be known at compile time.

lib/THC/THCTensorInfo.cuh
```diff
@@ -223,6 +233,23 @@ TensorInfo<T, IndexType>::collapseDims(int excludeDim) {
   return returnDim;
 }
 
+template <typename T, typename IndexType>
+TensorInfo<T, IndexType>
+TensorInfo<T, IndexType>::newNarrow(int dimension, IndexType firstIndex, IndexType size) const {
+  IndexType sz[MAX_CUTORCH_DIMS];
```
wickedfoo: This would now be a device function as well, and MAX_CUTORCH_DIMS is so large now too, I think that this would either be assigned to many registers or to local memory, negatively impacting perf.

Probably what needs to happen is add `int MaxDims = MAX_CUTORCH_DIMS` as an additional template argument to TensorInfo, and then specialize the kernel to restrict # of dims to 2 - 4 or something for the parallel version.

The excessive size of MAX_CUTORCH_DIMS would help your # of kernel launch problems as well.

Maybe it's only worthwhile to call the all-in-one kernel in the case that all of the tensors being cat'ed are contiguous and are dim-4 or less, so you don't need a stride array for any of them?

lib/THC/THCTensorMath.cuh
wickedfoo: re: constant memory, you could be calling this concurrently from different streams, so you need a different region of constant memory for each stream.

instead of global memory, i think i'd just specialize the kernel further to only handle contiguous inputs of 4 dims or less, so you don't need the stride array at all, and you don't need MAX_DIMS sizes.

wickedfoo: (and don't use TensorInfo perhaps, or you'd have to add the size of the size/stride array as a template parameter, and/or split TensorInfo into a base which is just size only)

killeent:
Updated the diff with some changes:

- Instead of passing TensorInfo structs for the to-be-cat'd tensors, we pass raw pointers to the data and a stride buffer of stride values. We then re-use the existing dimension size, and result tensor sizes, to calculate offsets for the to-be-cat'd tensors. This allows us to pass up to 128 tensors at a time, as opposed to 16 using the old approach
- Add template specializations for 1, 2D and contiguous input tensors, to speed up linearIndex -> offset mappings

Some Benchmarks:

Old Code Path:
`Averages of	10	trials used to generate results	
----------------------------------------------------------------	
CatArray for	2	1D Tensors took	4.7445297241211e-05	seconds	
CatArray for	4	1D Tensors took	5.0187110900879e-05	seconds	
CatArray for	8	1D Tensors took	7.932186126709e-05	seconds	
CatArray for	16	1D Tensors took	0.00016393661499023	seconds	
CatArray for	32	1D Tensors took	0.00027034282684326	seconds	
CatArray for	64	1D Tensors took	0.00049459934234619	seconds	
CatArray for	128	1D Tensors took	0.00099062919616699	seconds	
CatArray for	256	1D Tensors took	0.0018555879592896	seconds	
CatArray for	512	1D Tensors took	0.0039491891860962	seconds	
CatArray for	1024	1D Tensors took	0.0078642129898071	seconds	
CatArray for	2048	1D Tensors took	0.015282130241394	seconds	
CatArray for	4096	1D Tensors took	0.033820652961731	seconds	
CatArray for	8192	1D Tensors took	0.063122916221619	seconds	
----------------------------------------------------------------	
CatArray for	2	2D Tensors along dim=1 took	4.0292739868164e-05	seconds	
CatArray for	2	2D Tensors along dim=2 took	4.727840423584e-05	seconds	
CatArray for	4	2D Tensors along dim=1 took	7.7486038208008e-05	seconds	
CatArray for	4	2D Tensors along dim=2 took	5.3644180297852e-05	seconds	
CatArray for	8	2D Tensors along dim=1 took	8.3112716674805e-05	seconds	
CatArray for	8	2D Tensors along dim=2 took	8.5854530334473e-05	seconds	
CatArray for	16	2D Tensors along dim=1 took	0.00015788078308105	seconds	
CatArray for	16	2D Tensors along dim=2 took	0.00015175342559814	seconds	
CatArray for	32	2D Tensors along dim=1 took	0.0002760648727417	seconds	
CatArray for	32	2D Tensors along dim=2 took	0.00026459693908691	seconds	
CatArray for	64	2D Tensors along dim=1 took	0.00051555633544922	seconds	
CatArray for	64	2D Tensors along dim=2 took	0.00053470134735107	seconds	
CatArray for	128	2D Tensors along dim=1 took	0.0010017395019531	seconds	
CatArray for	128	2D Tensors along dim=2 took	0.0010006904602051	seconds	
CatArray for	256	2D Tensors along dim=1 took	0.0020315885543823	seconds	
CatArray for	256	2D Tensors along dim=2 took	0.0020285606384277	seconds	
CatArray for	512	2D Tensors along dim=1 took	0.0040197372436523	seconds	
CatArray for	512	2D Tensors along dim=2 took	0.004431414604187	seconds	
CatArray for	1024	2D Tensors along dim=1 took	0.0078263759613037	seconds	
CatArray for	1024	2D Tensors along dim=2 took	0.012330675125122	seconds	
CatArray for	2048	2D Tensors along dim=1 took	0.03656861782074	seconds	
CatArray for	2048	2D Tensors along dim=2 took	0.029906320571899	seconds	
CatArray for	4096	2D Tensors along dim=1 took	0.033309888839722	seconds	
CatArray for	4096	2D Tensors along dim=2 took	0.031117582321167	seconds	
CatArray for	8192	2D Tensors along dim=1 took	0.060645318031311	seconds	
CatArray for	8192	2D Tensors along dim=2 took	0.062959051132202	seconds	
----------------------------------------------------------------	
CatArray for	2	3D Tensors along dim=1 took	4.7540664672852e-05	seconds	
CatArray for	2	3D Tensors along dim=2 took	0.00012071132659912	seconds	
CatArray for	2	3D Tensors along dim=3 took	4.9233436584473e-05	seconds	
CatArray for	4	3D Tensors along dim=1 took	6.568431854248e-05	seconds	
CatArray for	4	3D Tensors along dim=2 took	9.2697143554688e-05	seconds	
CatArray for	4	3D Tensors along dim=3 took	5.8293342590332e-05	seconds	
CatArray for	8	3D Tensors along dim=1 took	8.7976455688477e-05	seconds	
CatArray for	8	3D Tensors along dim=2 took	0.00012013912200928	seconds	
CatArray for	8	3D Tensors along dim=3 took	0.00012049674987793	seconds	
CatArray for	16	3D Tensors along dim=1 took	0.00024595260620117	seconds	
CatArray for	16	3D Tensors along dim=2 took	0.00016570091247559	seconds	
CatArray for	16	3D Tensors along dim=3 took	0.00020656585693359	seconds	
CatArray for	32	3D Tensors along dim=1 took	0.00030195713043213	seconds	
CatArray for	32	3D Tensors along dim=2 took	0.00031402111053467	seconds	
CatArray for	32	3D Tensors along dim=3 took	0.00038647651672363	seconds	
CatArray for	64	3D Tensors along dim=1 took	0.00066125392913818	seconds	
CatArray for	64	3D Tensors along dim=2 took	0.00074398517608643	seconds	
CatArray for	64	3D Tensors along dim=3 took	0.00067653656005859	seconds	
CatArray for	128	3D Tensors along dim=1 took	0.0014249801635742	seconds	
CatArray for	128	3D Tensors along dim=2 took	0.0013843059539795	seconds	
CatArray for	128	3D Tensors along dim=3 took	0.0014357566833496	seconds	
CatArray for	256	3D Tensors along dim=1 took	0.0024216890335083	seconds	
CatArray for	256	3D Tensors along dim=2 took	0.0024879932403564	seconds	
CatArray for	256	3D Tensors along dim=3 took	0.0024538516998291	seconds	
CatArray for	512	3D Tensors along dim=1 took	0.0528568983078	seconds	
CatArray for	512	3D Tensors along dim=2 took	0.0045055866241455	seconds	
CatArray for	512	3D Tensors along dim=3 took	0.0044718027114868	seconds	
CatArray for	1024	3D Tensors along dim=1 took	0.031960463523865	seconds	
CatArray for	1024	3D Tensors along dim=2 took	0.0080471038818359	seconds	
CatArray for	1024	3D Tensors along dim=3 took	0.027758288383484	seconds	
CatArray for	2048	3D Tensors along dim=1 took	0.015772366523743	seconds	
CatArray for	2048	3D Tensors along dim=2 took	0.031630563735962	seconds	
CatArray for	2048	3D Tensors along dim=3 took	0.016491961479187	seconds	
CatArray for	4096	3D Tensors along dim=1 took	0.055450701713562	seconds	
CatArray for	4096	3D Tensors along dim=2 took	0.054601669311523	seconds	
CatArray for	4096	3D Tensors along dim=3 took	0.056299877166748	seconds	
CatArray for	8192	3D Tensors along dim=1 took	0.065966153144836	seconds	
CatArray for	8192	3D Tensors along dim=2 took	0.094370603561401	seconds	
CatArray for	8192	3D Tensors along dim=3 took	0.11704707145691	seconds`

pavanky:
@killeent Any chance you can add another column in the benchmarks comparing the time it takes to clone the final output array ?

killeent:
@pavanky what do you mean specifically? e.g.

`x = torch.cat({...}, dim)
  y = x:clone()`

?

pavanky:
@killeent I am assuming your earlier numbers were for `x = torch.cat({...}, dim)`. It would be interesting to see what performance you are getting as a percentage of peak expected performance. The easiest way to do this would be to compare with `x:clone()` after performing `cat.

killeent:
@pavanky https://gist.github.com/killeent/723ca46f4153ee15087ea9fd5929983b is for calling clone immediately after the cat call, for the new kernel implementation.

pavanky:
The clone times on size 8192 do not look right:

```
CatArray for	8192	1D Tensors took	0.0041048049926758	seconds. Clone:	0.059316992759705
```

CatArray of 8192 tensors can not be 14x faster than cloning a single array for the same size.

Any chance you can share the benchmark script ?

killeent:
https://gist.github.com/killeent/bcd13c48d5b894ce109ae90100004c17

pavanky:
@killeent you are not calling synchronize in the benchmark script. Without it it is not guaranteed that the operation is finished.

killeent:
Alright, this is now in a reviewable state. Changes since our last update:

- Utilize global memory to pass Tensor metadata for the input tensors to the kernel. We use per-stream scratch space for this memory, which is allocated once and re-used throughout the lifetime of the kernel
- Add further constraints on when we call this kernel - such as restricting it to non-empty tensors, and constraining it to operate on tensors of 4 dimensions or less
- Update unit tests to more accurately trigger the batching conditions

I used two scripts to benchmark:
- https://gist.github.com/killeent/e12e2d51a4cb533359867f05a5f3abeb
- https://gist.github.com/killeent/33476c9eb3a84354026f1bd0c190e4e6

I verified the timing for the first benchmark script (on small tensors) by using the cuda event timing scheme described here: https://devblogs.nvidia.com/parallelforall/how-implement-performance-metrics-cuda-cc/, and by running the script in nvprof and looking at the results. The 
'ms: #####" lines in the output are from the event timing.

I rely on the torch level CPU timer for the second benchmark script.

Old Scheme, first benchmark script: https://gist.github.com/killeent/a81943cbe3766f67b5a40dd6e71898fb
New Scheme, first benchmark script:
https://gist.github.com/killeent/b6477929fdaab21bfb97595a9252e6d4

After the initial cost of allocating the scratch space, we see speedups of 8-20x.

Old Scheme, comprehensive benchmark script: https://gist.github.com/killeent/ef32ed451a362d6cf6bcb837cb1e3010
New Scheme, comprehensive benchmark script:
https://gist.github.com/killeent/d5fe5d3bc239902fc9e31faf5969c8cc

Here we see speedups increase as the the number of input tensors increase (from roughly on par with the existing implementation when small, to pretty significant speedups when large. This extends to running thousands of cats on the same tensor input count/size.

killeent:
cc @soumith @pavanky @wickedfoo this can be reviewed now, I think

pavanky:
@killeent only one small change in the benchmarking scripts. You'll need to synchronize before starting the timer too. This ensures all previous cuda calls are completed by this point (in your case it ensures all the random number generation calls are completed).

killeent:
With cutorch.synchronize() called after RNG for Tensor initialization but before timer start in both benchmarking scripts:

Existing Kernel: https://gist.github.com/killeent/7ae26695c5227dd15e4fd9b6eee9fe1c, https://gist.github.com/killeent/b0b158fb9e49668517d0d36d1b04bc82
New Implementation: https://gist.github.com/killeent/7808d4c452d7edbce2d36e458ec0530c, https://gist.github.com/killeent/b818877a774190507dfcca2cc21bee82

lib/THC/THCTensorMath.cuh
```diff
@@ -23,4 +23,74 @@ __global__ void THCTensor_copyToDiagonal(T* a, T* b, ptrdiff_t start, ptrdiff_t
...
+#define CAT_ARRAY_MAX_INPUT_DIMS 4
+
+// Similar to any other IndexToOffset calculation for copying along a given dimension.
+template <typename IndexType, int Dims>
+struct CatArrIndexToOffset {
+  static inline __device__ IndexType compute(
+      const IndexType outputSize[Dims], const IndexType outputStride[Dims], const IndexType dimSize, const unsigned int concatDim, IndexType linearIndex) {
+    IndexType offset = 0;
+
+    for (int i = Dims - 1; i >= 0; --i) {
```
wickedfoo: the compiler should do it anyways, but stick a `#pragma unroll` on this

lib/THC/THCTensorMath.cuh
```diff
@@ -23,4 +23,74 @@ __global__ void THCTensor_copyToDiagonal(T* a, T* b, ptrdiff_t start, ptrdiff_t
...
+  T* data = inputs[blockIdx.y].input;
+  IndexType offset = inputs[blockIdx.y].offset;
+  IndexType dimSize = inputs[blockIdx.y].dimSize;
+  IndexType nElements = inputs[blockIdx.y].nElements;
+  IndexType dataOffset = offset * dimStride;
+
+  for (IndexType linearIndex = blockIdx.x * blockDim.x + threadIdx.x;
+      linearIndex < nElements;
+      linearIndex += gridDim.x * blockDim.x) {
+    IndexType elementOffset = CatArrIndexToOffset<IndexType, Dims>::compute(os.outputSize, os.outputStride, dimSize, concatDim, linearIndex);
```
wickedfoo: can you break this line so it doesn't wrap? keep under 80-100 columns

lib/THC/THCTensorTypeUtils.cu
```diff
@@ -112,12 +112,39 @@ TensorUtils<TENSOR_TYPE>::isContiguous(THCState* state,                 \
   return TENSOR_TYPE##_isContiguous(state, t);                          \
 }                                                                       \
                                                                         \
+bool                                                                    \
+TensorUtils<TENSOR_TYPE>::allContiguous(THCState* state,                \
+                                        TENSOR_TYPE** inputs,           \
+                                        int numInputs) {                \
+  assert(numInputs > 0);                                                \
```
wickedfoo: use `THAssert` instead

lib/THC/THCTensorTypeUtils.cu
```diff
@@ -112,12 +112,39 @@ TensorUtils<TENSOR_TYPE>::isContiguous(THCState* state,                 \
...
 TensorUtils<TENSOR_TYPE>::getDevice(THCState* state,                    \
                                     TENSOR_TYPE* t) {                   \
   return TENSOR_TYPE##_getDevice(state, t);                             \
 }                                                                       \
                                                                         \
+bool                                                                    \
+TensorUtils<TENSOR_TYPE>::allSameDevice(THCState* state,                \
+                                        TENSOR_TYPE** inputs,           \
+                                        int numInputs) {                \
+  assert(numInputs > 0);                                                \
```
wickedfoo: use `THAssert` instead

test/test.lua
```diff
@@ -3688,6 +3688,55 @@ function test.catArray()
    end
 end
 
+-- designed to specifically hit the batched kernel for catArray
```
wickedfoo: maybe have a test where the output is non-contiguous to begin with (since you support that I think?)

killeent:
Addressed @wickedfoo's comments, if its good @soumith can you merge? Thanks!

lib/THC/THCTensorMath.cuh
```diff
@@ -23,4 +23,80 @@ __global__ void THCTensor_copyToDiagonal(T* a, T* b, ptrdiff_t start, ptrdiff_t
...
+    IndexType offset = 0;
+
+#pragma unroll
+    for (int i = Dims - 1; i >= 0; --i) {
+      IndexType curDimSize = i == concatDim ? dimSize : outputSize[i];
+      IndexType curDimIndex = linearIndex % curDimSize;
+      IndexType curDimOffset = curDimIndex * outputStride[i];
+      offset += curDimOffset;
+
+      linearIndex /= curDimSize;
```
pavanky: Since both `/` and `%` are being used (and are costly), It may be a little bit more efficient to do something like this to avoid one of them:
```
IndexType nextDimIndex = linearIndex / currDimSize;
IndexType currDimIndex = linearIndex - currDimSize * nextDimIndex;
IndexType currDimOffset = currDimIndex * outputStride[1];
offset += currDimOffset;
linearIndex = nextDimIndex;
```

pavanky: I am also not sure if there is any point for the loop to go down to 0. Unless I am missing something, It should theoretically be sufficient to go down to 1 and return `offset + linearIndex * outputStride[0]` which would behave like a special case when `Dims == 1`

PR #637, opened 2016-12-15
Fixing various tests

pavanky:
- Increased the number of elements being used by distributions.
- Fixed indexFill to generate a number that can be used by all types.


soumith:
thanks!

Issue #636, opened 2016-12-15
multinomial_with_replacement fails intermittently.


pavanky:
This happens when the test suite is ran multiple times with the following error:

```
multinomial_without_replacement
sampled an index twice
BOOL violation condition=false
```

PR #638, opened 2016-12-15
Bugfix for multinomial distribution

pavanky:
- Ensures the index of the first bin from the cdf is returned.

- potential fix for #636


pavanky:
I am not 100% sure this is the right fix, but the tests passed for multiple runs.

soumith:
cc: @wickedfoo

pavanky:
The test mentioned in #636  still fails after running a few more times. However, the frequency at which it fails seems to have decreased.

wickedfoo:
Seems legit and makes sense to me, why not :)

PR #640, opened 2016-12-16
Revert "Bugfix of type in THCTensor macro."

soumith:
Reverts torch/cutorch#639


soumith:
this breaks symbol naming. I didn't realize it until @apaszke pointed to me, looked like a harmless PR.

PR #641, opened 2016-12-19
Add support for cremainder, cfmod

killeent:
As title. Things pertinent for review:

- Returning NAN from the OPs, this seems to work but I am not sure if this is the best way to do so. It mimics the behavior of TH
- Change up the isEqual function to properly handle Tensors with NAN values

Test Plan:
- Add unit tests


JoostvDoorn:
This breaks the install, see #643

soumith:
@killeent needs some CUDA 70 guards maybe?

Issue #643, opened 2016-12-20
error: no suitable conversion function from "__half" to "double" exists


ChijunZhang:
When building cutorch,
[ 15%] Building NVCC (Device) object lib/THC/CMakeFiles/THC.dir//./THC_generated_THCTensorMathPairwise.cu.o
/tmp/luarocks_cutorch-scm-1-8088/cutorch/lib/THC/THCTensorMathPointwise.cuh(457): error: operand types are incompatible ("__half" and "float")

/tmp/luarocks_cutorch-scm-1-8088/cutorch/lib/THC/THCTensorMathPointwise.cuh(467): error: operand types are incompatible ("__half" and "float")

/tmp/luarocks_cutorch-scm-1-8088/cutorch/lib/THC/THCTensorMathPointwise.cuh(514): error: no suitable conversion function from "__half" to "double" exists

3 errors detected in the compilation of "/tmp/tmpxft_00000df4_00000000-7_THCTensorMath2.cpp1.ii".
CMake Error at THC_generated_THCTensorMath2.cu.o.cmake:267 (message):
  Error generating file

my cuda version: cuda8.0
cudnn version:cuDNN5.1

xgastaldi:
Same issue here

soumith:
i fixed this by reverting the PR. things should work now. sorry for that.

ChijunZhang:
Thanks a lot

PR #646, opened 2016-12-21
cremainder, cfmod implementations (take 2)

killeent:
Same as #641, but should fix the build issues in #643.

Tested on CUDA 7.5, with half-precision support, as well.


soumith:
thanks!

PR #652, opened 2016-12-27
Make multinomial return a LongTensor (compatible with CPU version)

apaszke:
This makes the signature compatible with CPU version, but is a breaking change.

Unfortunately I had to remove an optimized implementation used for sampling a single value without any allocations. It depended on calling `normal` on the result tensor, but it won't work with long outputs.


soumith:
thanks!

PR #663, opened 2017-01-09
half<->float conversion cleanup

borisfom:
Using half conversions from TH.
Depends on https://github.com/torch/torch7/pull/901


borisfom:
Replaced by https://github.com/torch/cutorch/pull/680

PR #664, opened 2017-01-09
Instantiate 128kb of scratch space in GPU memory per-device by default

killeent:
For things such as the new CatArray kernel, it will be helpful to have a larger buffer of GPU scratch space available to serve as reusable GPU memory, without requiring use of the caching allocator. This diff bumps the default scratch space per device from 16 bytes to 128 kb.

cc @wickedfoo

### DIFF
--- a/lib/THC/THCGeneral.c
+++ b/lib/THC/THCGeneral.c
@@ -10,7 +10,12 @@
 #include <stdint.h>
 
 /* Size of scratch space available in global memory per each SM + stream */
-#define GLOBAL_SCRATCH_SPACE_PER_SM_STREAM 4 * sizeof(float)
+#define MIN_GLOBAL_SCRATCH_SPACE_PER_SM_STREAM 4 * sizeof(float)
+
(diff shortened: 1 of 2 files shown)


soumith:
looking forward, do we see a regime in which we dont use the caching allocator? I dont think so.
If this scratch space only helps catArray, seems wasteful. I'll accept it and merge it in, but can you rethink the reason for the scratch space.

soumith:
for example, the scratch space we need for batchedGemm, we just allocate on the fly, and we assume that caching allocator is used.

wickedfoo:
I think cutorch core algorithms should not depend upon what kind of allocator is being used, and potentially suffer bad performance if the allocator is not the caching allocator.

The cutorch library might be used embedded in other systems that are managing the GPU memory on top; e.g., caffe2 or other systems. Or, the tensor library can be a drop-in C/C++ replacement for N-dimensional GPU array manipulations elsewhere.

soumith:
mergefornow!

PR #670, opened 2017-01-11
Fix resize out of memory errors

### LINKED ISSUE #669: Resize gives out of memory when there is still free memory
It seems when resizing large Tensors out of memory errors occur relatively quickly. The example below is rather superficial, but extends to varying sequence lengths and batch sizes in RNNs when implemented with resize. Since resize does not provide any guarantee for the contents it should be able to allocate the needed memory (up to about n=8 in the example below).
```
1
7868/8104 Mbytes
2
6844/8104 Mbytes
3
5820/8104 Mbytes
4
4796/8104 Mbytes
5
3772/8104 Mbytes
THCudaCheck FAIL file=/tmp/luarocks_cutorch-scm-1-9879/cutorch/lib/THC/generic/THCStorage.cu line=66 error=2 : out of memory
```
```lua
require 'cutorch'
local a = torch.Tensor(1):cuda()
for n=1,100 do
   print(n)
   cutorch.synchronize()
   local freeMemory, totalMemory = cutorch.getMemoryUsage()
   print(string.format("%d/%d Mbytes", freeMemory/1024/1024, totalMemory/1024/1024))
   a:resize(1024, 262144, n)
end
```
This is not a new issue, I have checked with a version from about 9 months ago.

JoostvDoorn:
This fixes #669.

This changes resize behaviour for cases when memory is 'low'. If upon resize malloc fails, the current memory is freed first and another attempt at malloc is performed. The data in the tensor is then lost. Feel free to re-implement this in a better way.

### DIFF
--- a/lib/THC/generic/THCStorage.cu
+++ b/lib/THC/generic/THCStorage.cu
@@ -61,7 +61,21 @@ void THCStorage_(resize)(THCState *state, THCStorage *self, ptrdiff_t size)
                                  size * sizeof(real),
                                  THCState_getCurrentStream(state));
     if(err != cudaSuccess) {
-      THCHeapUpdate(state, -size * sizeof(real));
+      if(self->flag & TH_STORAGE_FREEMEM) {
+         // Try reallocating memory on failure
+         THCudaCheck(
+           (*self->allocator->free)(self->allocatorContext, self->data));
+         THCHeapUpdate(state, -self->size * sizeof(real));
+         self->data = NULL;
+         err = (*self->allocator->malloc)(self->allocatorContext,
+                                    (void**)&(data),
+                                    size * sizeof(real),
+                                    THCState_getCurrentStream(state));
+      }
+                                
+      if(err != cudaSuccess) {
+         THCHeapUpdate(state, -size * sizeof(real));
+      }
     }
     THCudaCheck(err);
 


lib/THC/generic/THCStorage.cu
```diff
@@ -61,7 +61,21 @@ void THCStorage_(resize)(THCState *state, THCStorage *self, ptrdiff_t size)
                                  size * sizeof(real),
                                  THCState_getCurrentStream(state));
     if(err != cudaSuccess) {
-      THCHeapUpdate(state, -size * sizeof(real));
+      if(self->flag & TH_STORAGE_FREEMEM) {
+         // Try reallocating memory on failure
+         THCudaCheck(
+           (*self->allocator->free)(self->allocatorContext, self->data));
+         THCHeapUpdate(state, -self->size * sizeof(real));
```
apaszke: Shouldn't this have a positive delta?

lib/THC/generic/THCStorage.cu
```diff
@@ -61,7 +61,21 @@ void THCStorage_(resize)(THCState *state, THCStorage *self, ptrdiff_t size)
...
+         self->data = NULL;
```
apaszke: I think it would be better to reallocate the previous size if this bigger allocation still fails. The docs say that after resize tensor content is unspecified so it's ok, but having NULL data pointer seems dangerous.

apaszke: Or allocate a single value. Then we can say that the state of the storage is unspecified but it's still a valid object.

apaszke:
Also this is a no-op if the selected allocator implements realloc. You can find an alternative branch above.

lib/THC/generic/THCStorage.cu
JoostvDoorn: This first part frees the memory so the heap gets smaller. (this is the same as line 46)

JoostvDoorn:
I do not see the alternative branch?

apaszke:
Lines 26-40.

JoostvDoorn:
Right, I guess it would be better. I saw realloc was removed in dce6cd89aa78d308dd4360484ac21da3fc80c9d0 by @soumith, do you know why?

apaszke:
Probably because having realloc that does free + malloc is useless (and it doesn't conform to standard semantics, but I'm not sure if we require the existing memory to be copied).

killeent:
@apaszke what do you mean by useless in this context?

lib/THC/generic/THCStorage.cu
killeent: It seems fine to me to leave this as is - if the allocation fails the resulting storage being invalid seems reasonable to me. It's going to trigger the _THError anyway.

apaszke:
I just meant that it's not what realloc does and you can simulate that in other parts of code if you really need that fallback.

JoostvDoorn:
I agree that having a realloc that does not follow standard semantics (c specifications) for realloc is not useful.

soumith:
after much debate, this is not going in. @colesbury thinks people (unfortunately) rely on the behavior that resize = realloc. I dont want to break userland.
If one needs a workaround, they have to delete the tensor and reallocate a new one.

JoostvDoorn:
Thanks for getting back to me @soumith. Moving forward will you consider pull requests that fix this issue in specific modules? E.g. in cudnn RNN for reserve.

soumith:
yes, happy to merge your patches int cudnn RNN or elsewhere that does explicit delete / create Tensors.

PR #675, opened 2017-01-13
Ensure atomicAdd(double) is visible to host side code

pavanky:
Just replicating behavior of the cuda headers


pavanky:
Note to self: Always do full builds before sending in patches upstream.

soumith:
i'm glad you caught it before me :)

PR #678, opened 2017-01-16
Refactoring of THCNumerics

borisfom:
Here, this refactoring was initially driven by idea of supporting both pseudo and native fp16 for a single arch. That support proved to be unfeasible, but the refactoring made the code much cleaner and shorter, take a look. 
Also, CUDA_HALF_INSTRUCTIONS compile-time switch is turned on for 6.0 and 6.2+ architectures in this version - if you encounter any new issues with accuracy on Pascal, try turning it off.


borisfom:
Replaced by https://github.com/torch/cutorch/pull/679

PR #680, opened 2017-01-17
Using half conversions from TH

borisfom:
Using half conversions from TH.
Depends on torch/torch7#901


borisfom:
I think this request just fell under the cracks - very simple change improving runtime efficiency (no memcpy, hey) and eliminating code duplication.

soumith:
thanks!

PR #683, opened 2017-01-25
Reduced type set for Android

borisfom:
@soumith: Here, new take on reduced type set,  much less invasive.
Also, addresses the issue of versioned libraries.


borisfom:
Verified that reduced set allows cunn test for float pass.

borisfom:
In this version, only math functions for types other than 'float' and 'long' are being cut.
The rest instantiated as usual. ~2x .so size drop.

PR #688, opened 2017-01-29
Compile with all asserts enabled by default

apaszke:
Otherwise, when an input is invalid (e.g. out of bound indices), some error are silent and may give corrupted outputs.


fmassa:
Are asserts in CUDA kernels widely available? I thought that it was not the case, and according to http://docs.nvidia.com/cuda/cuda-c-programming-guide/index.html#assertion it seems that it doesn't work on Mas OS

apaszke:
It compiles fine on my mbp. They're should have been no-ops, but actually they seemed to trigger and interrupt the kernel.

pavanky:
> **Assertions are for debugging purposes**. They can affect performance and it is therefore recommended to disable them in production code. They can be disabled at compile time by defining the NDEBUG preprocessor macro before including assert.h. Note that expression should not be an expression with side effects (something like (++i > 0), for example), otherwise disabling the assertion will affect the functionality of the code.

This PR removes out of bounds checks and replaces them with asserts. This should not exist in production code. This PR should not have been merged.

apaszke:
Are you sure that passing asserts add any more overhead than the comparison? I think the note is that they **potentially** can be expensive. @ngimel do you know what's the overhead when they don't trigger?

pavanky:
@apaszke That is practically how every CUDA program works. You are not finding a new technique that improves the CUDA performance magically.

soumith:
there seems to be real value to adding them, as long as perf is not affected down to death. The kernels where these assertions can possibly be triggered are pretty bandwidth-bound anyways.

The number of user reports of illegal memory access because they went out of bounds is ridiculous and only parallels our getParameters bug reports.

apaszke:
I'm sorry? I never said it will improve performance. I'm just saying that it shouldn't add overhead under normal operation, but it will trigger a clear error if you have out of bound indices. Right now, if you have invalid values in the index tensor, no values will be written to the output, leaving nans and other garbage there - and it will be a completely silent error.

pavanky:
@soumith 

This is even worse than I thought. `NDEBUG` is being passed as a pre-processor macro and the docs say this:

```
They can be disabled at compile time by defining the NDEBUG preprocessor macro before including assert.h.
```

So there are **no bounds checks happening** at all when built in release mode.

apaszke:
No, it's not. The thing that this PR adds is filtering out `-DNDEBUG` from the release mode flags.

soumith:
i'm confused. i thought @apaszke verified that scraping these flags in Release mode does add bounds checks...

pavanky:
nvm, I see that NDEBUG is replaced. But

apaszke:
Yeah I did verify. They get pretty loud once you pass in an invalid input 😄

pavanky:
20% overhead for a simple kernel:
https://gist.github.com/pavanky/e431d61862315e08be1ddda992c2cfe7

apaszke:
`(506.268311 / 481.653076 - 1) * 100 = 5,11%`, not 20%. And no silent UB.

pavanky:
@apaszke I edited already :)

soumith:
if 5% is what we have to eat on our assertable kernels like ClassNLLCriterion and index*, we should eat that perf. These kernels are never the bottleneck in the pipeline. Seems reasonable.

pavanky:
@apaszke I jumped the gun by not completely reading your comments earlier. I am sorry about that.

@soumith 

That sounds reasonable.

But a few smaller considerations:

1) The asserts are completely destroying the state when they fail. You'll have to do `cudaDeviceReset` to recover. From the docs:

> No more commands can be sent to this device until cudaDeviceReset() is called to reinitialize the device.

2) Each thread that is failing is printing the assertion failure to `stderr`.

apaszke:
Hmmm, that's awful. In most cases they're supposed to be lethal anyway, but we should be able to recover from this in all other ones 😕

ngimel:
@soumith, assert overhead used to be pretty large (even when it was not triggered), apparently it is smaller now according to @pavanky measurements. I'm not familiar with index* code, but what is assert giving you that subsequent cudaGetLastError does not? Pros for assert: exact location of the failure, cons: perf penalty. As for not being able to recover after assert, same is true for illegal memory access reported by cudaGetLastError so one is no better than the other.

soumith:
@ngimel we cant call cudaGetLastError for index bounds checks, for example in ClassNLLCriterion here: https://github.com/torch/cunn/blob/master/lib/THCUNN/ClassNLLCriterion.cu#L52 (unless i am mistaken)

apaszke:
@ngimel The problem is that invalid indexing won't necessarily result in invalid memory accesses.

ngimel:
@soumith, @apaszke I see (btw https://github.com/torch/cunn/blob/master/lib/THCUNN/ClassNLLCriterion.cu#L51 lacks (int) cast). I don't see a good solution :-(

PR #694, opened 2017-02-06
Adding support for integer type tensors using CudaHostAllocator

pavanky:


pavanky:
@soumith I hope the for loop isn't an ugly solution :) I just did not want to repeat the same few lines over and over.

soumith:
this is great, thanks!

PR #695, opened 2017-02-06
Static build support + Query CUDA driver, runtime versions

pavanky:


lib/THC/CMakeLists.txt
```diff
@@ -3,6 +3,7 @@ CMAKE_POLICY(VERSION 2.8)
 
 SET(CMAKE_MODULE_PATH ${CMAKE_CURRENT_SOURCE_DIR}/cmake ${CMAKE_MODULE_PATH})
 
+SET(CUDA_ATTACH_VS_BUILD_RULE_TO_CUDA_FILE OFF)
```
soumith: you use Visual Studio? :P

pavanky: No, but apparently the default option was making the `.cu` files compile multiple times (4x more?) back when this was added.

PR #699, opened 2017-02-10
Bitwise operations

pavanky:
- lshift, rshift, bitand, bitor, bitxor
- Adds matching functionality from https://github.com/torch/torch7/pull/933


soumith:
thanks!

PR #702, opened 2017-02-14
Improve THCCachingHostAllocator performance by making it reclaim less aggressively

gchanan:


soumith:
@gchanan just fyi this PR breaks on gcc 4.4.7 which only has support for c++0x. I dont care if we want to keep this c++0x-only requirement, maybe it's okay we move to c++11

gchanan:
@soumith it's probably not worth it to remove support for gcc 4.4.7 just for this change.  I had assumed that emplace existed since emplace_back existed for std::vector (and those functions are in the same standards proposals), but it looks like gcc didn't implement them until later: https://gcc.gnu.org/bugzilla/show_bug.cgi?id=44436.

gchanan:
https://github.com/torch/cutorch/pull/703

PR #703, opened 2017-02-15
Fix gcc 4.4.7 build.

gchanan:


soumith:
auto also isn't supported afaik

soumith:
at least it was giving me an error

gchanan:
it was giving you an error because emplace doesn't exist, so it doesn't know the type for auto.

gchanan:
e.g. https://github.com/torch/cutorch/commit/eb3fe1904d18934f177972ad89c973be92b97dc2 is the original THCCachingHostAllocator commit has auto.

soumith:
got it. thanks!

PR #704, opened 2017-02-17
WIP: CuTorch Mode

killeent:
This diff implements mode for CudaTensors, which is currently not a supported function. This first stab implementation sets up the scaffolding for the call, and also has a correct (if inefficient) implementation.

Next steps:

- [x] 1. Refine serial implementation so that it has less allocs/frees during the loop. Instead use a buffer - so we don't need to rely on the Caching Allocator for good perf.

- [x] 2. Extend to all types.

- [x] 3. Clean up code. Wrap calls outside of Lua, and verify no memory leaks when using Cuda functions.

- [x] 4. Benchmarking Script.

- [x] 5. Write a custom kernel that is more efficient by fusing the operations so that we can compute the mode/indices in parallel.


lib/THC/THCTensorMode.cu
```diff
@@ -0,0 +1,202 @@
...
+  THCudaTensor* temp = src;
+  for (int i = 0; i < THLongStorage_size(position); ++i) {
+    THCudaTensor_select(state, self, temp, 0, THLongStorage_data(position)[i]);
+    temp = self;
+  }
+}
+
+THC_API void THCudaTensor_calculateMode(THCState* state,
+                                        THCudaTensor* values,
+                                        THCudaLongTensor* indices,
```
killeent: @soumith I think this should just be a CPU Tensor for the indices, but I wasn't sure. What do you think?

lib/THC/THCTensorMode.cuh
```diff
@@ -0,0 +1,282 @@
...
+  if (tidx == 0) {
+    mode = smem[imem[0]];
+  }
+  __syncthreads(); // broadcast mode
+
+  // Finally, we need to find the "an" index of the mode in the input Tensor. The API does
+  // not constrain which index we pick, so it can be any of the indices that contain the mode.
+  // We will do a reduction to find the index. First, we mark indices that are equal to the mode,
+  // i.e bmem[i] = true if input[i] == mode
+  if (tidx < sliceSize) {
```
wickedfoo: Consider using `reduceBlock` twice, or adding functionality to `reduceBlock` where you can provide multiple values instead of duplicating more reduction code; e.g.,

```
int indices[2] = { ... };
reduceBlock<int, Min<int>, 2>(imem, sliceSize, indices, Min<int>, MAX_INT);
```

all you need is a min reduction on an integer index value, where the value is the current `tidx` or `stidx` for threads that have a mode value, and MAX_INT otherwise.

Then, if we add the warp shuffle optimization to `reduceBlock`, this code will get faster as well.

lib/THC/THCTensorMode.cuh
```diff
@@ -0,0 +1,282 @@
...
+
+template <>
+struct BinaryAddOp<unsigned int> {
+  __host__ __device__ inline unsigned int operator()(const unsigned int a, const unsigned int b) {
+    return a + b;
+  }
+};
+
+template <typename T, class BinaryOp, int Power2ScanSize>
+__device__ void segmentedInclusivePrefixScan(T *smem, bool *bmem, BinaryOp binop) {
```
wickedfoo: move these segmented scans into the THCReduceApplyUtils so they can be usable elsewhere (a prefix scan is a kind of reduction, it gives all of the partial reductions of a sequence)

lib/THC/THCTensorMode.cuh
```diff
@@ -0,0 +1,282 @@
...
+  //      imem = [4, 1, 4, 3, 4, 5]
+  //
+  // Loop 3 (offset = 1)
+  //
+  // (0, 1) --> cmem[1] == 1 <= cmem[0] = 2, do nothing
+  //
+  // So at the end at cmem[0] we have the maximum count = 2, and the
+  // corresponding index = 4
+#pragma unroll
+  for (unsigned int offset = Power2Size / 2; offset > 0; offset >>= 1) {
```
wickedfoo: This is more duplicated reduction code, would be nice to restructure to use helper functions like `reduceBlock`.

The issue is that you have different buffers declared for using shared memory, but in this case you are reducing what is effectively a:

```
struct ModeCount { int count; int index; };
```

with maximum count. 

`smem` is used with `bmem`, and `cmem` depends upon `bmem`. 
`imem` is used together with `cmem`.

All of them do not need to be resident at the same point in time. This is decreasing occupancy as well. In fact, at most you need (max(sizeof(T), sizeof(unsigned int)) + sizeof(unsigned int)) * Power2Size shared memory. You can dynamically allocate this instead, and repurpose the memory at each step.

Then, you can view that memory as an array of my `struct ModeCount` above, and just use `reduceBlock` on that.

lib/THC/THCTensorMode.cuh
```diff
@@ -0,0 +1,282 @@
...
+  }
+  __syncthreads(); // barrier for bmem initialization
+
+  // Next, we initialize another shared memory buffer cmem which will be used in the
+  // Segmented Prefix Sum
+  __shared__ unsigned int  cmem[Power2Size];
+
+  // We set cmem to be the negation of bmem. In particular, we can think of cmem[i] = true
+  // iff A[i-1] == A[i] in our original sorted slice.
+  cmem[tidx] = !bmem[tidx];
```
wickedfoo: feels like these can be initialized directly from `smem` instead, no need to go through `bmem`.

lib/THC/THCTensorMode.cuh
```diff
@@ -0,0 +1,282 @@
...
+  // We re-use the bmem buffer for this computation. In particular, we can think of
+  // B[i] = true indicating the start of a sequence of equal values in the sorted
+  // list.
+
+  if (tidx == 0) {
+    bmem[tidx] = true;  // for setting element 0
+  }
+
+  // Compares elements (0, 1), (2, 3), ... and sets 1, 3, ...
+  bmem[tidx * 2 + 1] = THCNumerics<T>::ne(smem[tidx * 2], smem[tidx * 2 + 1]); // (0, 1), (1, 2), etc.
```
wickedfoo: these seem to be potential access violations to me, because max `tidx` is `blockDim.x - 1` which is `Power2Size - 1`, and you're thus indexing `Power2Size` in the last thread which is invalid.

It may not be showing up in `cuda-memcheck` because you have so many different regions of shared memory arrays which are in one contiguous space, thus indexing over the bounds results in still hitting a valid region of shared memory, or because it always allocates in multiples of 16 bytes or something.

If you try and coalesce together the regions of shared memory using dynamic allocation instead (see below), you might see the problem in `cuda-memcheck`.

Also, this feels unnecessary; you can just initialize `cmem` directly from `smem` below.

lib/THC/THCTensorMode.cuh
killeent: will do

lib/THC/THCTensorMode.cuh
killeent: What part specifically feels unnecessary? I'm a little confused here: do you mean doing the following:

```
// assuming I fix the initialization
bmem[tidx * 2 + 1] = THCNumerics<T>::ne(smem[tidx * 2], smem[tidx * 2 + 1]);
cmem[tidx * 2 + 1] = !bmem[tidx * 2 + 1];
```

Which would I think save on a __syncthreads() barrier?

killeent: Also I'm not sure that I'm doing an out-of-bounds access here: In the kernel `Power2Size` is always `2 * blockDim.x` given our kernel launch parameters.

lib/THC/THCTensorMode.cuh
killeent: Note that given this current implementation, we actually need:
```
Power2Size * (sizeof(T) + (2 * sizeof(unsigned int)))
```

because we refer back to the sorted input to find the mode.

lib/THC/THCTensorMode.cuh
wickedfoo: Ah nevermind about the access violation, sorry.

Yeah, saving on the syncthreads. While the compiler may not attempt to store and then re-load from shared memory (since it is not marked as `volatile`), I'd be deliberate about having the value in a register:

```
bool flag = THCNumerics<T>:: ...
bmem[tidx * 2 + 1] = flag;
cmem[tidx * 2 + 1] = !flag;
````

**Comment by killeent** [comment]
I have addressed @wickedfoo's comments:

- Move from SoA to AoS during reductions / segmentedInclusivePrefixScan
- Re-use existing generic reduction code for reductions. Required adding a new generic reduction kernel with per-thread register reduction prior to warp reduction
- More efficient uses of shared memory for increased occupancy
- Remove unnecessary use of `__syncthreads()` during segmented scan initialization

File: lib/THC/THCReduceApplyUtils.cuh [hunk 7]
```diff
@@ -72,6 +72,32 @@ __device__ T reduceBlock(T* smem,
...
+
+#pragma unroll
+  for (int i = 1; i < N; ++i) {
+    ++offset;
+    T next = offset < numVals ? threadVals[i] : init;
+    local = reduceOp(local, next);
+  }
+
+  // Needed if threadVals points to the same shared memory as smem
+  if (SyncAfterLocalReduction) {
```
wickedfoo: I think the idea is that threadVals should be in registers, not smem

lib/THC/THCReduceApplyUtils.cuh
```diff
@@ -72,6 +72,32 @@ __device__ T reduceBlock(T* smem,
...
+// Block-wide reduction where each thread locally reduces N
+// values before letting a single warp take over
+template <typename T, typename ReduceOp, int N, bool SyncAfterLocalReduction>
+__device__ T reduceBlockN(T *smem,
+                         T threadVals[N],
+                         int numVals,
+                         ReduceOp reduceOp,
+                         T init) {
+  int offset = threadIdx.x * N;
+  T local = offset < numVals ? threadVals[0] : init;
```
wickedfoo: An alternative could be assuming that numVals is exactly N * blockDim.x, so:

```
T local = threadVals[0];
#pragma unroll
for (int i = 1; i < N; ++i) {
  local = reduceOp(local, threadVals[i]);
}
```

assuming that out of bounds threads initialize all threadVals[i] to the identity element for the reduction

lib/THC/THCReduceApplyUtils.cuh
wickedfoo: I think the idea is that threadVals is always in registers, not smem, thus this is not needed

(just as `T threadVal` for `reduceBlock` is always in a register)

lib/THC/THCTensorMode.cuh
```diff
@@ -0,0 +1,292 @@
...
+    Power2Size>(
+        ubpmem,
+        SegmentedScanOp<struct ModeUnsignedBoolPair, BinaryAddOp<unsigned int> >(BinaryAddOp<unsigned int>()));
+  // assumes scan syncs at the end
+
+  // Next, we reinterpret the ubpmem buffer as pairs of unsigned integers (i.e. we treat the
+  // boolean flag regions as integers). We initialize these to represent indices, and we'll call
+  // this buffer I
+  struct ModeUnsignedPair *uupmem = reinterpret_cast<struct ModeUnsignedPair *>(ubpmem);
+  uupmem[tidx].index = tidx;
```
wickedfoo: Why store this in shared memory? The point of reduceBlockN is that the input data can be held in registers, as you're immediately going to reduce N values intra-thread:

```
ModeUnsignedPair uup[2];
uup[0].index = tids;
uup[0].val = ...;
uup[1].index = stidx;
uup[1].val = ...;

educeBlockN<struct ModeUnsignedPair, MaxReduceOp<struct ModeUnsignedPair>, 2, true>(uupmem, uup, ...);
```

lib/THC/THCTensorMode.cuh
```diff
@@ -0,0 +1,292 @@
...
+  }
+  __syncthreads(); // broadcast mode
+
+  // Finally, we need to find the "an" index of the mode in the input Tensor. The API does
+  // not constrain which index we pick, so it can be any of the indices that contain the mode.
+  // We will do a reduction to find the index. We go back to using the (index, flag) buffer
+  // arrangment. First, we mark indices that are equal to the mode, i.e B[i] = true if
+  // input[i] == mode, and initialize C[i] to be the index
+  if (tidx < sliceSize) {
+    ubpmem[tidx].flag = THCNumerics<T>::eq(input[linearOffset + tidx], mode);
```
wickedfoo: Same as above: why store this in shared memory? The point of reduceBlockN is that the input data can be held in registers

lib/THC/THCTensorMode.cuh
```diff
@@ -0,0 +1,292 @@
...
+    ubpmem[stidx].val = stidx;
+  }
+  __syncthreads(); // barrier for initialization of ubpmem
+
+  // Then we perform a similar reduction to the one above, except this time we update
+  // the element if the element at the base position is not equal to the mode and
+  // the element at the offset position is. At the end, C[0] will contain an index
+  // with the mode.
+  struct ModeUnsignedBoolPair match = {0, false};
+  match = reduceBlockN<struct ModeUnsignedBoolPair, MatchReduceOp<struct ModeUnsignedBoolPair>, 2, true>
```
wickedfoo: Feels like this can be merged with the previous reduction, so that there is just a single reduction.

Reduce the tuple (value, count, index) by max count, return `value` as mode and `index` as the index. This would be the last index among the mode elements, though.

wickedfoo:
The only thing I really care about here is that reduceBlockN should take input values in registers, not smem. Merging the two reductions or not I leave up to you.

killeent:
Address more comments:

- Reduce in registers prior to performing block-wide reductions
- Remove reduceBlockN option for data initially in shared memory

PR #709, opened 2017-02-27
Fix bug where pinned memory event could be recorded on incorrect device

colesbury:
Previously, the device corresponding to the stream in block.streams would be ambiguous for default (NULL) streams. The event could be recored on an incorrect device.

This changes THCStream so that we have non-NULL THCStream* objects for default streams. The underlying `cudaStream_t` object is still NULL. I've added a test for this case in test.lua.  

I also changed the handling of `THCStream` to use smart pointers so that they're freed whenever a block is freed. Previously, the could be leaked in `emptyCache()`. I'm using `set` and `shared_ptr` instead of `unordered_set` and `unique_ptr` because of poor C++11 support in GCC 4.4.


colesbury:
cc @gchanan

lib/THC/THCStream.c
```diff
@@ -13,18 +16,38 @@ THCStream* THCStream_new(int flags)
   return self;
 }
 
+THC_API THCStream* THCStream_defaultStream(int device)
+{
+  // default streams aren't refcounted
+  THAssert(device >= 0 && device < MAX_DEVICES);
+  THCStream* stream = &default_streams[device];
+  stream->device = device;
```
gchanan: I don't know if this would ever practically go wrong, but my reading of the C standard is it's not technically safe.

test/test.lua
```diff
@@ -4213,6 +4213,31 @@ if os.getenv('THC_CACHING_ALLOCATOR') == '1' then
...
+
+   function test.cachedPinnedMemoryMultiGPU()
+      local device_count = cutorch.getDeviceCount()
+      if device_count < 2 then
+         return
+      end
+
+      local cyclesPerMs = getCyclesPerMs()
+      local t = cutorch.createCudaHostTensor(1)
+      ptr = t:data()
```
colesbury: Oops... I've been writing too much Python

PR #710, opened 2017-02-27
Expose a mutex that is held around cudaFree() calls.

colesbury:
NCCL can deadlock if `cudaFree()` is called while it's launching kernels.
This exposes a mutex that can be held to prevent `cudaFree()` calls in the
caching allocator.


pavanky:
@colesbury @soumith Not sure if including a C++ header in a public header is a good idea.

PR #719, opened 2017-03-03
Fixes to cat 

### LINKED ISSUE #717: torch.cat fails when dimension not specified
torch.cat fails with illegal memory access error when the dimension is not specified.
```
require('cutorch')

local a = torch.CudaTensor(1, 5)
local b = torch.CudaTensor(1, 5)

local c = torch.cat(a, b, 2) -- Works
cutorch.synchronize()
print('torch.cat(a,b,2) passed')

local d = torch.cat(a, b) -- fails
cutorch.synchronize()
print('torch.cat(a,b) passed')
```

Error:

```
$ th /tmp/cat.lua 
torch.cat(a,b,2) passed	
THCudaCheck FAIL file=/home/pyalamanchili/Workspace/torch/cutorch/init.c line=246 error=77 : an illegal memory access was encountered
/home/pyalamanchili/Workspace/torch/local/bin/lua: /tmp/cat.lua:11: cuda runtime error (77) : an illegal memory access was encountered at /home/pyalamanchili/Workspace/torch/cutorch/init.c:246
stack traceback:
	[C]: in function 'synchronize'
	/tmp/cat.lua:11: in main chunk
	[C]: in function 'dofile'
	...kspace/torch/local/lib/luarocks/rocks/trepl/scm-1/bin/th:150: in main chunk
	[C]: in ?
```

The CPU version (i.e. replacing `torch.CudaTensor` with `torch.Tensor`) works.

pavanky:
- Fixes #717

### DIFF
--- a/lib/THC/generic/THCTensorMath.cu
+++ b/lib/THC/generic/THCTensorMath.cu
@@ -223,7 +223,7 @@ void THCTensor_(catArray)(THCState *state, THCTensor *result,
         // update offset
         offset += dimSize;
       }
-      cudaMemcpy(d_inputs, stackInputs, j * sizeof(CatArrInputTensor<real, unsigned int>), cudaMemcpyHostToDevice);
+      THCudaCheck(cudaMemcpy(d_inputs, stackInputs, j * sizeof(CatArrInputTensor<real, unsigned int>), cudaMemcpyHostToDevice));
 
       // Next, let's consider how we set our kernel launch parameters.
       // We borrow from THCApply, which the kernel's internal indexing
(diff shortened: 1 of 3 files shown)


lib/THC/generic/THCTensorMath.cu
```diff
@@ -203,15 +203,15 @@ void THCTensor_(catArray)(THCState *state, THCTensor *result,
 
     // Template Declarations for dim = 1, 2, 3, 4
 #define HANDLE_CASE(DIMS) \
-  CatArrayBatchedCopy<real, unsigned int, DIMS><<<applyGrid, applyBlock>>>(data, d_inputs, param, ldimension, param.outputStride[dimension]);
+  CatArrayBatchedCopy<real, unsigned int, DIMS><<<applyGrid, applyBlock>>>(data, d_inputs, param, cat_dimension, param.outputStride[cat_dimension]);
```
pavanky: This was the problematic line. It should have used `outputStride[ldimension]`. Changed the variable name to be more explicit so it can be better tracked.

soumith:
thanks!

PR #720, opened 2017-03-03
Support half precision in baddbmm

guillaumekln:
As discussed in [cunn/#417](https://github.com/torch/cunn/issues/417), this change adds the support of half precision in `baddbmm` by iterating over each batch.


soumith:
thanks!

Issue #707, opened 2017-02-26
Caching allocator (enabled by default now?) causes cuda OOM


vadimkantorov:
I've installed a fresh Torch on a new machine recently. My code started to OOM.

I have a function that allocates a few thousand medium-sized local CudaTensors (4Gb worth of memory). nil'ing the table with these tensors and calling `collectgarbage` didn't help to free the memory.

I do not have `THC_CACHING_ALLOCATOR` set to 1, but I gave it a try by explicitly setting `export THC_CACHING_ALLOCATOR=0`. It solved the problem. It seems the caching allocator is enabled by default now.

Is there a way to flush the allocator's cache from Torch? (I guess it'd be a binding for https://github.com/torch/cutorch/blob/master/lib/THC/THCCachingAllocator.cpp#L172). For now I'll be disabling the allocator for my code.

The fact that the code OOMs suggests the allocator doesn't handle well near-OOM situation or that there are some allocations (inside cublas?) that bypass the caching allocator.

colesbury:
Do you have a repro script? I know the cuDNN calls we use allocate device memory outside the caching allocator, but I didn't think cuBLAS did.

vadimkantorov:
I will check if my code still blows with the caching allocator. I vaguely remember the crash was inside some matrix multiply code, but I'll let you know when I check it.

A manual forced release of unused blocks may still be useful for situations of using external libraries like cuDNN, and maybe some nasty allocation patterns.

marcel1991:
How do you disable the caching allocator? Where do you set the flag to false? 

I have the same problem that after a cutorch update my memory usage is now higher and I can only use a smaller batch size compared to before.

soumith:
you set it in your terminal before you run torch:

```
export THC_CACHING_ALLOCATOR=0
th foo.lua
```

marcel1991:
Okay thank you! That solved the OOM problem.

soumith:
do you have a small script that can reproduce your issue? that'll help us fix this once and for all.

marcel1991:
No just a huge one :/ If you are talking to me.

vadimkantorov:
For me back then it happened, when I allocated a big list of tensors, each of whose were of different size (like `#rois x 128`). And when this list went out of range and new list was allocated this same memory was not well reused.

PR #722, opened 2017-03-06
Enable or disable caching allocator from Lua

guillaumekln:
The CUDA caching allocator is now enabled by default but is know to consume more memory (#707). It can be controlled with the `THC_CACHING_ALLOCATOR` environment variable but some applications may want to enforce a value to mitigate their memory usage without user interaction.

Here I propose a simple approach to disable or re-enable the CUDA caching allocator from Lua code.

### DIFF
--- a/init.c
+++ b/init.c
@@ -1,8 +1,6 @@
 #include "utils.h"
 #include "luaT.h"
 #include "THCGeneral.h"
-#include "THCCachingAllocator.h"
-#include "THCCachingHostAllocator.h"
 #include "THCSleep.h"
 #include "THCTensorRandom.h"
 #include "THCHalf.h" // for CUDA_HALF_TENSOR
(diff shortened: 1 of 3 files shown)


gchanan:
this doesn't really seem worth the complexity; maybe I'm missing the use case, but it seems semi-far fetched that someone is going to use a caching allocator, realize they are using too much memory, and turn off the caching allocator at run time, over just not using the caching allocator to begin with.

And it's not clear this would even solve the above problem; for example, if you use the caching allocator and haven't freed some allocations when you switch, you'd have to empty the cache again after you free, otherwise you only mark the cached blocks as unallocated and don't actually return them to the system.

guillaumekln:
The primary use case is to disable the caching allocator from the client application, typically during its initialization. I agree the current implementation is somewhat naive and certainly does not properly handle the switch at runtime (i.e. after some allocations have already occur).

However, I believe applications should be able to enforce a value as the caching allocator can require as much as 2GB more memory. Currently, I have to repeat to users to use this environment flag to fix their out of memory issues.

Let me know if there is a better way to implement this feature. Or if you don't want to support such feature, that's fine too and you can close this PR.

gchanan:
what about just having an accessor/acccessors and using that in the client to warn/error about the specific client not being used with caching?

guillaumekln:
That's sound good to me. I will re-open a PR with this approach.

Thanks!

PR #723, opened 2017-03-08
add implementation of inclusive scan via upsweep-downsweep

killeent:
A slightly more efficient Scan that uses upsweep/downsweep like mechanisms.

Tested outside of cutorch codebase on buffers of size 2, 3, 21, 32, 33, 64 and verified that it properly calculated the prefix sum when templatized via an addition binary op.

### DIFF
--- a/lib/THC/THCScanUtils.cuh
+++ b/lib/THC/THCScanUtils.cuh
@@ -5,6 +5,50 @@
 
 // Collection of in-kernel scan / prefix sum utilities
 
+// Inclusive Scan via an upsweep/downsweep mechanism. Assumes:
+//
+// 1. Power2ScanSize is a power of 2. This code still works for collections that
+// do not exactly contain a power of 2 number of elements, simply round up to the
+// nearest power of 2 and then call.
+//
+// 2. That there are two-elements per thread, i.e. the size of the smem storage
+// is 2 * blockDim.x * sizeof(T).
+//
+// Consider a (+)-Scan on the following elements:
+//
+// Upsweep:
+//
+//    0  1  2  3  4  5  6  7
+//       1     5     9    13
+//             6          22
+//                        28
+//
+// Downsweep:
+//                  15
+//         3     10    21
+template <typename T, class BinaryOp, int Power2ScanSize>
+__device__ void inclusivePrefixScan(T *smem, BinaryOp binop) {
+  // Reduce step ("upsweep")
+#pragma unroll
+  for (int stride = 1; stride < Power2ScanSize; stride <<= 1) {
+    int index = (threadIdx.x + 1) * stride * 2 - 1;
+    if (index < Power2ScanSize) {
+      smem[index] = binop(smem[index], smem[index - stride]);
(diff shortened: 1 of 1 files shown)


lib/THC/THCScanUtils.cuh
```diff
@@ -5,6 +5,50 @@
...
+// Downsweep:
+//                  15
+//         3     10    21
+template <typename T, class BinaryOp, int Power2ScanSize>
+__device__ void inclusivePrefixScan(T *smem, BinaryOp binop) {
+  // Reduce step ("upsweep")
+#pragma unroll
+  for (int stride = 1; stride < Power2ScanSize; stride <<= 1) {
+    int index = (threadIdx.x + 1) * stride * 2 - 1;
+    if (index < Power2ScanSize) {
```
wickedfoo: "This code still works for collections that
 +// do not exactly contain a power of 2 number of elements, simply round up to the
 +// nearest power of 2 and then call."

This is not true here, you should pass in an extra size parameter for the data instead?

pavanky:
Can you note the performance improvements ?

lib/THC/THCScanUtils.cuh
```diff
@@ -5,6 +5,50 @@
...
+//    0  1  2  3  4  5  6  7
+//       1     5     9    13
+//             6          22
+//                        28
+//
+// Downsweep:
+//                  15
+//         3     10    21
+template <typename T, class BinaryOp, int Power2ScanSize>
+__device__ void inclusivePrefixScan(T *smem, BinaryOp binop) {
```
wickedfoo: Also this function should take an input and pass an output like the others, instead of assuming the values are already in shared memory?

lib/THC/THCScanUtils.cuh
killeent: I guess to clarify, the algorithm works as long as you have Power2ScanSize space in smem, but yes we could add a size parameter to condition on as well.

lib/THC/THCScanUtils.cuh
killeent: In the other case each thread is responsible for a single element, where as in this case each thread has two associated elements. So we could match it via:

`__device__ void inclusivePrefixScan(T *smem, T a, T b, T *out, BinaryOp op`

but I think it is a little less clean than compared with the others.

lib/THC/THCScanUtils.cuh
wickedfoo: actually, nevermind, it won't work with a non-power of 2 size, this is fine.

wickedfoo: duh, I take that back, it will work. A size parameter sounds like a good addition, since not every scan will involve a power-of-2 size, especially the tail end of a set of data (either that, or the smem will have to be reset with an identity value for the reduction; e.g., for +, would have to be filled with 0).

soumith:
@pavanky these changes are all for the upcoming `mode` kernel that @killeent is working on.

PR #724, opened 2017-03-08
Add CUDA caching allocator accessor

guillaumekln:
This PR allows clients to check whether the caching allocator is enabled or not to inform users about expected performance and/or memory usage.

It is a partial version of https://github.com/torch/cutorch/pull/722 with a unique getter on the memory allocator.


soumith:
thanks!

PR #739, opened 2017-03-30
Get rid of warp-synchronous code

ngimel:
Time to get rid of warp-synchronous code - it is unsafe.


wickedfoo:
looks good to me

are there finer-grained synchronization primitives coming (warp-wide barriers)?

ngimel:
Yes, but no details yet.

soumith:
thanks!

killeent:
Great. As @wickedfoo has previously mentioned, I had to disable these to get any sort of useful output from the racecheck tool.

PR #742, opened 2017-04-04
Use thrust::inclusive_scan for 1D cumsum/cumprod

colesbury:
For large 1D tensors thrust::inclusive_scan is much faster than our current implementation.

For 4 million elements thrust takes about 430µs vs 200ms for the current implementation.

### DIFF
--- a/lib/THC/generic/THCTensorMathScan.cu
+++ b/lib/THC/generic/THCTensorMathScan.cu
@@ -2,6 +2,27 @@
 #define THC_GENERIC_FILE "generic/THCTensorMathScan.cu"
 #else
 
+#ifndef THC_REAL_IS_HALF
+template<class BinaryFunction>
+__host__ void THCTensor_(scanThrust)(
+    THCState *state,
+    THCTensor *dst,
+    THCTensor *src,
+    BinaryFunction binary_op)
+{
+  THCThrustAllocator thrustAlloc(state);
+  thrust::device_ptr<real> src_data(THCTensor_(data)(state, src));
+  thrust::device_ptr<real> dst_data(THCTensor_(data)(state, dst));
+  ptrdiff_t size = THCTensor_(nElement)(state, src);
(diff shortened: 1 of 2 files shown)


pavanky:
May be consider using thrust + for loop for scanning along inner most dimension if the current version is really that slow.

lib/THC/generic/THCTensorMathScan.cu
```diff
@@ -57,12 +78,22 @@ template<class BinaryFunction>
...
+  THArgCheck(dimension >= 0 && dimension < ndim, 3, "dimension %d out of range",
+      dimension + TH_INDEX_BASE);
 
+  THCTensor_(resizeAs)(state, self_, src);
   THCTensor *self = THCTensor_(newContiguous)(state, self_);
   src = THCTensor_(newContiguous)(state, src);
 
-  if (dimension == THCTensor_(nDimension)(state, src) - 1) {
+#ifndef THC_REAL_IS_HALF
+  if (ndim == 1) {
```
fmassa: don't we need to check that the tensors are contiguous? Or they are supposed to be contiguous at this point?

fmassa: Just saw line 88, forget about it

colesbury:
@pavanky, there are cases where the current implementation would be faster for 2D+ tensors so I didn't want to completely switch to that (if the outer dimension is large enough to parallelize over). If the 2D+ case is a bottleneck we may want to revisit this and possibly pull in segmented scan from moderngpu.

wickedfoo:
This is the general tradeoff that much of cutorch makes. Many cutorch functions operate under the assumption that there is a large enough batch (you are doing something along dimension D of an N-dimensional tensor) rather than one thing along a 1-d array. The implementation usually just targets the first (batch) case, leaving bad performance for the non-batch case, which requires a different specialization.

PR #744, opened 2017-04-07
Generic TopK implementation

killeent:
This PR makes TopK generic. The main changes required to do this were:

1. Extend the float --> unsigned integer bijective function to all other Tensor types
2. Make it possible to act on data types whose underlying representation requires 64bits


killeent:
@soumith @wickedfoo This is ready.

lib/THC/THCTensorTopK.cuh
```diff
@@ -0,0 +1,470 @@
...
+    return v - 128;
+  }
+};
+
+template <>
+struct TopKTypeConfig<short> {
+  typedef unsigned int RadixType;
+
+  static inline __device__ RadixType convert(short v) {
+    return 32768u + v;
```
wickedfoo: add a static assert that sizeof(short) == 2 to ensure that this is a correct constant

lib/THC/THCTensorTopK.cuh
```diff
@@ -0,0 +1,470 @@
...
+    return v - 32768;
+  }
+};
+
+template <>
+struct TopKTypeConfig<int> {
+  typedef unsigned int RadixType;
+
+  static inline __device__ RadixType convert(int v) {
+    return 2147483648u + v;
```
wickedfoo: add a static assert that sizeof(int) == 4 to ensure that this is a correct constant

lib/THC/THCTensorTopK.cuh
```diff
@@ -0,0 +1,470 @@
...
+    return v - 2147483648u;
+  }
+};
+
+template <>
+struct TopKTypeConfig<long> {
+  typedef unsigned long long int RadixType;
+
+  static inline __device__ RadixType convert(long v) {
+    return 9223372036854775808ull + v;
```
wickedfoo: add a static assert that sizeof(long) == 8 to ensure that this is a correct constant

PR #745, opened 2017-04-11
block wide reduction with multiple values to reduce at once

killeent:
This extends reduceBlock so that each thread does N reductions at once. For example, the N=2 reduction of:

(1, 2)
(3, 4)
(5, 6)
(7, 8)

yields (16, 20)

### DIFF
--- a/lib/THC/THCReduceApplyUtils.cuh
+++ b/lib/THC/THCReduceApplyUtils.cuh
@@ -19,59 +19,6 @@ __device__ __forceinline__ IndexType getLinearBlockId() {
     blockIdx.x;
 }
 
-// Block-wide reduction in shared memory helper; only threadIdx.x == 0 will
-// return the reduced value
-template <typename T, typename ReduceOp>
-__device__ T reduceBlock(T* smem,
-                         int numVals,
-                         T threadVal,
-                         ReduceOp reduceOp,
-                         T init) {
-  if (numVals == 0) {
-    return init;
-  }
-
-  if (threadIdx.x < numVals) {
-    smem[threadIdx.x] = threadVal;
-  }
-
-  // First warp will perform reductions across warps
-  __syncthreads();
-  if ((threadIdx.x / warpSize) == 0) {
-    T r = threadIdx.x < numVals ? smem[threadIdx.x] : init;
-
-    for (int i = warpSize + threadIdx.x; i < numVals; i += warpSize) {
-      r = reduceOp(r, smem[i]);
-    }
-
-    smem[threadIdx.x] = r;
-  }
-
-  // First thread will perform reductions across the block
-  __syncthreads();
-
-  T r = init;
-  if (threadIdx.x == 0) {
-    r = smem[0];
-
-    int numLanesParticipating = min(numVals, warpSize);
-
-    if (numLanesParticipating == 32) {
-      // Unroll for warpSize == 32 and numVals >= 32
-#pragma unroll
-      for (int i = 1; i < 32; ++i) {
-        r = reduceOp(r, smem[i]);
-      }
-    } else {
-      for (int i = 1; i < numLanesParticipating; ++i) {
-        r = reduceOp(r, smem[i]);
-      }
-    }
-  }
-
-  return r;
-}
-
 // Return value is in threadVals for threadIdx.x == 0
 template <typename T, typename ReduceOp, int N>
 __device__ void reduceNBlock(T *smem,
@@ -141,6 +88,19 @@ __device__ void reduceNBlock(T *smem,
     }
   }
 }
+
+// Block-wide reduction in shared memory helper; only threadIdx.x == 0 will
+// return the reduced value
+template <typename T, typename ReduceOp>
+__device__ T reduceBlock(T* smem,
+                         int numVals,
+                         T threadVal,
+                         ReduceOp reduceOp,
+                         T init) {
+  reduceNBlock<T, ReduceOp, 1>(smem, &threadVal, numVals, reduceOp, init);
+  return threadVal;
+}
+
 
 // Block-wide reduction where each thread locally reduces N
 // values before letting a single warp take over - assumes
--- a/lib/THC/THCReduceApplyUtils.cuh
+++ b/lib/THC/THCReduceApplyUtils.cuh
@@ -21,7 +21,7 @@ __device__ __forceinline__ IndexType getLinearBlockId() {
 
 // Return value is in threadVals for threadIdx.x == 0
 template <typename T, typename ReduceOp, int N>
-__device__ void reduceNBlock(T *smem,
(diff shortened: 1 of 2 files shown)


lib/THC/THCReduceApplyUtils.cuh
```diff
@@ -72,6 +72,76 @@ __device__ T reduceBlock(T* smem,
   return r;
 }
 
+// Return value is in threadVals for threadIdx.x == 0
+template <typename T, typename ReduceOp, int N>
```
wickedfoo: can you make `reduceBlock` call this function with `N == 1`?

there's a lot of duplicated code in this file.
Similarly, `reduceNBlock` versus `reduceBlockN` is really confusing naming. Perhaps this version should be `reduceMultipleBlock`should be this version, whereas the multiple values per thread could be `reduceBlock`, named the same as the N == 1 case, since it's reducing all of them.

lib/THC/THCReduceApplyUtils.cuh
```diff
@@ -72,6 +72,76 @@ __device__ T reduceBlock(T* smem,
...
+
+  if (threadIdx.x < numVals) {
+#pragma unroll
+    for (int i = 0; i < N; ++i) {
+      smem[threadIdx.x * N + i] = threadVals[i];
+    }
+  }
+  __syncthreads();
+
+  if ((threadIdx.x / warpSize) == 0) {
```
wickedfoo: can we get the warp shuffle version based on CUDA_ARCH as part of this diff as well?

lib/THC/THCReduceApplyUtils.cuh
```diff
@@ -72,6 +72,76 @@ __device__ T reduceBlock(T* smem,
...
+    for (int i = 0; i < N; ++i) {
+      threadVals[i] = init;
+    }
+    return;
+  }
+
+  if (threadIdx.x < numVals) {
+#pragma unroll
+    for (int i = 0; i < N; ++i) {
+      smem[threadIdx.x * N + i] = threadVals[i];
```
wickedfoo: I think this would be slightly faster if the threads were writing contiguously:

`smem[i * numVals + threadIdx.x] = threadVals[i]`

lib/THC/THCReduceApplyUtils.cuh
```diff
@@ -19,64 +19,101 @@ __device__ __forceinline__ IndexType getLinearBlockId() {
...
+  // where to put the outputs of each of the n things we are reducing. If
+  // nLP = 32, then we have the 32 outputs for the first threadVal,
+  // followed by the 32 outputs for the second threadVal, etc.
+  int numLanesParticipating = min(numVals, warpSize);
+
   if ((threadIdx.x / warpSize) == 0) {
-    T r = threadIdx.x < numVals ? smem[threadIdx.x] : init;
+#pragma unroll
+    for (int i = 0; i < N; ++i) {
+      threadVals[i] = threadIdx.x < numVals ? smem[i * numVals + threadIdx.x] : init;
```
wickedfoo: actually, these values are already in `threadVals[i]`, so:
```
threadVals[i] = threadIdx.x < numVals ? threadVals[i] : init;
```

wickedfoo: actually, these values are already in `threadVals[i]`, so:
```
threadVals[i] = threadIdx.x < numVals ? threadVals[i] : init;
```

lib/THC/THCReduceApplyUtils.cuh
```diff
@@ -19,64 +19,101 @@ __device__ __forceinline__ IndexType getLinearBlockId() {
...
+#pragma unroll
+      for (int j = 0; j < N; ++j) {
+        threadVals[j] = reduceOp(threadVals[j], smem[j * numVals + i]);
+      }
     }
 
-    smem[threadIdx.x] = r;
+#pragma unroll
+    for (int i = 0; i < N; ++i) {
+      smem[i * numLanesParticipating + threadIdx.x] = threadVals[i];
```
wickedfoo: I think this is a potential bug in both the old and new code, if numVals < warpSize, then the smem might not be large enough.

This is a good reason to have a `__host__ __device size_t getSmemSize(int numVals)` for this code as well, packaged up as a struct. For the static `numVals` version, you can know that statically.

So it should probably be something like:
```
if (threadIdx.x < numLanesParticipating) {
#pragma unroll
  for (int i = 0; i < N; ++i) { ... }
}
```

lib/THC/THCReduceApplyUtils.cuh
```diff
@@ -19,64 +19,101 @@ __device__ __forceinline__ IndexType getLinearBlockId() {
...
-  // First thread will perform reductions across the block
   __syncthreads();
 
-  T r = init;
   if (threadIdx.x == 0) {
-    r = smem[0];
-
-    int numLanesParticipating = min(numVals, warpSize);
+#pragma unroll
+    for (int i = 0; i < N; ++i) {
```
wickedfoo: similarly, I think this little block isn't needed, the first lane in the first warp will already have these starter values in `threadVals`.

lib/THC/THCReduceApplyUtils.cuh
```diff
@@ -19,64 +19,101 @@ __device__ __forceinline__ IndexType getLinearBlockId() {
...
-// return the reduced value
-template <typename T, typename ReduceOp>
-__device__ T reduceBlock(T* smem,
-                         int numVals,
-                         T threadVal,
-                         ReduceOp reduceOp,
-                         T init) {
+// Return value is in threadVals for threadIdx.x == 0
+template <typename T, typename ReduceOp, int N>
+__device__ void reduceNValuesInBlock(T *smem,
```
wickedfoo: add a comment above that this helper is meant to reduce N different values concurrently, or something

lib/THC/THCReduceApplyUtils.cuh
killeent: So we basically need numVals * N * sizeof(T) space in shared memory.

Additionally, wouldn't the conditional below be:

`if (numLanesParticipating > warpSize && threadIdx.x < numLanesParticipating)`

Because if we have less than 32 values to reduce then we can proceed directly to the single-thread reduction below.

wickedfoo: numLanesParticipating is always <= warpSize, and this chunk of code is only executed by the first warp anyways.

But 
`if ((threadIdx.x / warpSize) == 0) {` up top could be changed to something like:

`if ((threadIdx.x / warpSize) == 0 && numVals > warpSize)`

so we avoid this block of code entirely if numVals < warpSize.

wickedfoo: Another alternative is add a guarantee that this code would only ever be called with at least one warp's worth of elements, and just remove these cases.

Issue #746, opened 2017-04-11
CudaCharTensor:abs() returning negative values


gchanan:
th> x=torch.CudaCharTensor(5)
                                                                      [0.0001s]
th> x
-16
  7
 64
  6
  0
[torch.CudaCharTensor of size 5]

                                                                      [0.0002s]
th> x:abs()
-16
  7
 64
  6
  0
[torch.CudaCharTensor of size 5]

PR #747, opened 2017-04-12
Fix abs with char and short cuda types.

gchanan:
Fixes https://github.com/torch/cutorch/issues/746.

I'm not really sure how the current code doesn't infinite loop (compiler optimizes it away?  It infinite loops for me if I try to printf in there).

There are also two abs functions defined: see: http://stackoverflow.com/questions/8226705/g-abs-on-a-short-int-appears-to-turn-it-into-a-double.

We want the int one; the template version doesn't compile since you get errors like: "calling a __host__ function("std::abs<short> ") from a __global__ function("kernelPointwiseApply2< ::Tensor_abs_Short_Op, short, short, unsigned int, (int)-1, (int)2> ") is not allowed"


soumith:
ha! good catch.

PR #751, opened 2017-04-18
THCReduce noncontigdim kernel improvements.

csarofeen:
THCReduce noncontigdim kernel improvements. Added extra kernel and heuristics to improve smaller tensor reductions.

### DIFF
--- a/lib/THC/THCReduce.cuh
+++ b/lib/THC/THCReduce.cuh
@@ -11,14 +11,58 @@
 #include "THCTensorTypeUtils.cuh"
 #include "THCReduceApplyUtils.cuh"
 
-// Threads per thread block
-#define THC_NONCONTIG_REDUCE_BLOCK_SIZE 32 * 16
+#include <cstdio>
 
-template <typename IndexType>
-__device__ __forceinline__ IndexType getReduceNoncontigDimSliceIndex() {
-  // Each thread handles one slice
-  return getLinearBlockId<IndexType>() * THC_NONCONTIG_REDUCE_BLOCK_SIZE + threadIdx.x;
+#define LOCAL_MAX_BLOCK_SIZE 512
+
+template <typename ModifyOp,
+          typename ReduceOp,
+          typename T,
+          typename IndexType,
+          int ADims, int BDims>
+__global__ void
+kernelReduceNoncontigDim_shared(TensorInfo<T, IndexType> out,
+                         TensorInfo<T, IndexType> in,
+                         IndexType reductionStride,
+                         IndexType reductionSize,
+                         IndexType totalSlices,
+                         T init,
+                         ModifyOp modifyOp,
+                         ReduceOp reduceOp) {
+  IndexType threadLane  = threadIdx.x;
+  IndexType groupID      = threadIdx.y;
+  IndexType sliceIndex  = blockIdx.x * blockDim.x + threadLane;
+  IndexType sliceStride = gridDim.x * blockDim.x;
+
+  __shared__ T local_reduce[LOCAL_MAX_BLOCK_SIZE];
+  //If multiple groups of threads work on the same reduction use shared memory
+  //Otherwise use a register for reduction as only one thread is doing reduction
+  T* shmem = &local_reduce[threadIdx.x + threadIdx.y * blockDim.x];
+
+  for(;sliceIndex<totalSlices; sliceIndex+=sliceStride){
+    //  while(sliceIndex < totalSlices){
+    *shmem = init;
+
+    IndexType outOffset = IndexToOffset<T, IndexType, ADims>::get(sliceIndex, out);
+    IndexType inOffset = IndexToOffset<T, IndexType, BDims>::get(sliceIndex, in);
+    IndexType stride = reductionStride * blockDim.y;
+
+    for(IndexType i=groupID; i<reductionSize; i+=blockDim.y){
+      (*shmem) = reduceOp(*shmem, modifyOp(in.data[inOffset]) );
+      inOffset += stride;
+    }
+
+    __syncthreads();
+
+    if(groupID == 0){
+      for(IndexType i=1; i<min( (IndexType) blockDim.y, reductionSize); i++){
+        *shmem = reduceOp(*shmem, *(shmem + blockDim.x * i) );
+      }
+      out.data[outOffset] = *shmem;
+    }
+  }
 }
+
 
 // Kernel that handles an entire reduction of a slice of a tensor per each thread
 template <typename ModifyOp,
@@ -26,9 +70,6 @@ template <typename ModifyOp,
           typename T,
           typename IndexType,
           int ADims, int BDims>
-#if __CUDA_ARCH__ >= 350
-__launch_bounds__(32 * 16, 4)
-#endif
 __global__ void
 kernelReduceNoncontigDim(TensorInfo<T, IndexType> out,
                          TensorInfo<T, IndexType> in,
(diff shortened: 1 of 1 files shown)


ngimel:
@soumith, @apaszke This would allow to switch to expand_as/sum (instead of addr/gemv) when adding bias in linear functions, with performance gains, esp for smaller linear sizes.

lib/THC/THCReduce.cuh
```diff
@@ -38,30 +79,22 @@ kernelReduceNoncontigDim(TensorInfo<T, IndexType> out,
...
-  T r = init;
+    for(IndexType i = 0; i<reductionSize; i++){
+      r = reduceOp(r, modifyOp(in.data[inOffset]) );
+      inOffset += reductionStride;
+    }
 
-  for (IndexType i = 0; i < reductionSize; ++i) {
-    r = reduceOp(r, modifyOp(in.data[inOffset]));
-    inOffset += reductionStride;
+    __syncthreads();
```
apaszke: Is this necessary?

lib/THC/THCReduce.cuh
```diff
@@ -200,11 +223,33 @@ bool THC_reduceDim(THCState* state,
...
-      return false;
+
+    //If there are a large number of outputs to the reduction, avoid syncthreads
+    //in kernel
+    if(outElements > 512*16){
+      block = dim3(512);
+    }else{
+      //x dim does different slices
+      //y dim helps with a slice
+      //If we only have 8 loops, don't bother sharing work across ydim
```
apaszke: Should it be 16 loops?

lib/THC/THCReduce.cuh
```diff
@@ -200,11 +223,33 @@ bool THC_reduceDim(THCState* state,
     block = getContigReduceBlock(outElements, reductionSize);
     smemSize = sizeof(typename TensorUtils<TensorType>::DataType) * block.x;
   } else {
-    if (!getNoncontigReduceGrid(outElements, grid)) {
-      return false;
+
+    //If there are a large number of outputs to the reduction, avoid syncthreads
```
apaszke: Both kernels have syncthreads right now

lib/THC/THCReduce.cuh
```diff
@@ -200,11 +223,33 @@ bool THC_reduceDim(THCState* state,
...
+      while(ydim > 1){
+        block.x /= 2;
+        block.y *= 2;
+        ydim /= 2;
+      }
+    }
+    long gridx  = THCCeilDiv( outElements, (long)block.x);
+    if (gridx > 1024){
+      long n_loops = THCCeilDiv(outElements, (long) (1024 * block.x) );
+      gridx = outElements / (block.x*n_loops);
```
apaszke: Are you sure this is ok? If you remove the ceil it is equivalent to setting gridx to 1024.

lib/THC/THCReduce.cuh
```diff
@@ -11,24 +11,65 @@
...
 
-// Threads per thread block
-#define THC_NONCONTIG_REDUCE_BLOCK_SIZE 32 * 16
+#include <cstdio>
 
-template <typename IndexType>
-__device__ __forceinline__ IndexType getReduceNoncontigDimSliceIndex() {
-  // Each thread handles one slice
-  return getLinearBlockId<IndexType>() * THC_NONCONTIG_REDUCE_BLOCK_SIZE + threadIdx.x;
+#define LOCAL_MAX_BLOCK_SIZE 512
```
apaszke: It seems that this constant is used for shared mem size, but is not used when computing the block size. Is that ok?

lib/THC/THCReduce.cuh
```diff
@@ -11,24 +11,65 @@
...
+    }
+
+    __syncthreads();
+
+    if(groupID == 0){
+      for(IndexType i=1; i<min( (IndexType) blockDim.y, reductionSize); i++){
+        *shmem = reduceOp(*shmem, *(shmem + blockDim.x * i) );
+      }
+      out.data[outOffset] = *shmem;
+    }
```
pavanky: Why is this just limited to `groupID == 0` ? Wouldn't reducing to half the groups at each step be faster ?

fmassa:
@apaszke I think @ngimel meant that we could use expand + add instead of `addr` in the forward of `Linear`, and sum the gradOutputs in the backward (instead of gemv). And it also avoids creating and resizing and filling with 1s at every iteration the `add_buffer`. I think this is related to [this discussion in the slack](https://pytorch.slackarchive.io/general/page-54/ts-1491516499928349)

apaszke:
@fmassa I know what's the deal with expand+add vs fill+addr, I'm just asking how is it related to this change. I don't know why I thought that expanded tensors are contiguous, nvm.

lib/THC/THCReduce.cuh
```diff
@@ -11,24 +11,65 @@
...
+__global__ void
+kernelReduceNoncontigDim_shared(TensorInfo<T, IndexType> out,
+                         TensorInfo<T, IndexType> in,
+                         IndexType reductionStride,
+                         IndexType reductionSize,
+                         IndexType totalSlices,
+                         T init,
+                         ModifyOp modifyOp,
+                         ReduceOp reduceOp) {
+  IndexType threadLane  = threadIdx.x;
```
killeent: threadLane seems like a bit of a misnomer here. I'm not sure how this corresponds to the lane in the warp.

lib/THC/THCReduce.cuh
```diff
@@ -11,24 +11,65 @@
...
+  IndexType groupID      = threadIdx.y;
+  IndexType sliceIndex  = blockIdx.x * blockDim.x + threadLane;
+  IndexType sliceStride = gridDim.x * blockDim.x;
```
killeent: Similarly, sliceStride is a bit confusing - this is actually the stride with which to get the next slice for reduction, but the variable name makes it sound like the stride for elements within a slice.

lib/THC/THCReduce.cuh
```diff
@@ -11,24 +11,65 @@
...
+  for(;sliceIndex<totalSlices; sliceIndex+=sliceStride){
+    //  while(sliceIndex < totalSlices){
+    *shmem = init;
+
+    IndexType outOffset = IndexToOffset<T, IndexType, ADims>::get(sliceIndex, out);
+    IndexType inOffset = IndexToOffset<T, IndexType, BDims>::get(sliceIndex, in);
+    IndexType stride = reductionStride * blockDim.y;
+
+    for(IndexType i=groupID; i<reductionSize; i+=blockDim.y){
+      (*shmem) = reduceOp(*shmem, modifyOp(in.data[inOffset]) );
```
killeent: I'm not sure exactly how this works. I could be wrong, but aren't we hitting shared memory every time here? If we want different threads in the "group" to reduce things in registers wouldn't we need a local variable?

lib/THC/THCReduce.cuh
csarofeen: https://github.com/csarofeen/cutorch/blob/master/lib/THC/THCReduce.cuh#L239-L244
Ensures block size = 512, was a little bit of a misnomer as I enforced 512 instead of having it as a max.

lib/THC/THCReduce.cuh
csarofeen: It is unfortunately necessary. We're trying to prevent warps from getting too far ahead which will have negative effects on the memory system.

lib/THC/THCReduce.cuh
csarofeen: Yes, I'll fix this comment.

lib/THC/THCReduce.cuh
csarofeen: Yes, I'll fix this comment.

lib/THC/THCReduce.cuh
csarofeen: I will review this again to make sure it is correct. It's mainly for load balancing the internal slice loop.

lib/THC/THCReduce.cuh
csarofeen: It might be, I could actually try as I forced blockdim.y to be a multiple of 2 so the logic shouldn't be too bad. Will check.

lib/THC/THCReduce.cuh
csarofeen: You're correct, was a remnant from when I was using a 1-D block. Will name it something more appropriate.

lib/THC/THCReduce.cuh
csarofeen: Do you have a suggestion on this name?

lib/THC/THCReduce.cuh
csarofeen: Will check, but compiler tends to optimize it to registers (this is why there's a shared mem volatile flag).

csarofeen:
https://gist.github.com/csarofeen/80e8e567d49e3a2511d6bcd7bd891a98
can be used for benchmarking linear improvements.
@apaszke Tensor is contiguous but reduction is on non-contiguous dimension.

lib/THC/THCReduce.cuh
apaszke: I know it enforces it, but I think it would be better to use the constant in both places. Otherwise there's no point in separating it from the code.

lib/THC/THCReduce.cuh
apaszke: How can they get too far from each other? If the ops have uneven branches?

csarofeen:
Still working on this.

PR #768, opened 2017-05-12
Add a nonContigDim reduction kernel to improve latency for small tensors

csarofeen:
Will improve performance of linear layers in pytorch.

### DIFF
--- a/lib/THC/THCReduce.cuh
+++ b/lib/THC/THCReduce.cuh
@@ -30,6 +30,99 @@ template <typename ModifyOp,
 __launch_bounds__(32 * 16, 4)
 #endif
 __global__ void
+kernelReduceNoncontigDim_shared(TensorInfo<T, IndexType> out,
+                         TensorInfo<T, IndexType> in,
+                         IndexType reductionStride,
+                         IndexType reductionSize,
+                         IndexType totalSlices,
+                         T init,
+                         ModifyOp modifyOp,
+                         ReduceOp reduceOp) {
+
+  IndexType sliceIndex  = blockIdx.x * blockDim.x + threadIdx.x;
+  IndexType sliceStride = gridDim.x * blockDim.x;
+
+  __shared__ T local_reduce[THC_NONCONTIG_REDUCE_BLOCK_SIZE];
+  T* shmem = &local_reduce[threadIdx.x + threadIdx.y * blockDim.x];
+  T load_reg[4];
+  T local_reg;
+
+  for(;sliceIndex<totalSlices; sliceIndex+=sliceStride){
+    local_reg = init;
+
+    const IndexType outOffset =
+      IndexToOffset<T, IndexType, ADims>::get(sliceIndex, out);
+    const IndexType inOffset =
+      IndexToOffset<T, IndexType, BDims>::get(sliceIndex, in);
+
+    //Unroll this loop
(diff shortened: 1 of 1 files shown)


lib/THC/THCReduce.cuh
```diff
@@ -30,6 +30,99 @@ template <typename ModifyOp,
...
+    const IndexType outOffset =
+      IndexToOffset<T, IndexType, ADims>::get(sliceIndex, out);
+    const IndexType inOffset =
+      IndexToOffset<T, IndexType, BDims>::get(sliceIndex, in);
+
+    //Unroll this loop
+    //for(IndexType i=threadIdx.y; i<reductionSize; i+=blockDim.y){
+    //  local_reg += in[inOffset + i * reductionStride];
+    //}
+    for(IndexType i=threadIdx.y; i<reductionSize; i+=blockDim.y*4){
```
apaszke: Wouldn't it be better to rewrite this for loop to only do full 4-element reductions and have a single if afterwards that optionally performs an incomplete step? Isn't that a more "classical" approach to loop unrolling? It'd save some comparisons inside the loop.

lib/THC/THCReduce.cuh
```diff
@@ -30,6 +30,99 @@ template <typename ModifyOp,
...
+    int dimy = blockDim.y;
+    while(dimy > 1){
+      __syncthreads();
+      if( threadIdx.y == 0 && (dimy%2 != 0) ){
+        *shmem = reduceOp(*shmem, *(shmem + (dimy-1) * blockDim.x) );
+      }
+      if(threadIdx.y < dimy/2){
+        *shmem = reduceOp(*shmem, *(shmem + (dimy/2)*blockDim.x) );
+      }
+      dimy /= 2;
```
apaszke: Wouldn't it be better to use the shuffling instructions here? It seems that cooperating groups of threads make up at most half of the warp and warp size is always divisible by group size.

lib/THC/THCReduce.cuh
csarofeen: This unrolling will have no divergence as warps will all take the same path. It won't improve performance.

lib/THC/THCReduce.cuh
csarofeen: I'm syncing across warps as cross warps are participating in the reduction (it's on threadIdx.y). It would make the code less readable as I'd have to transpose the block mapping to the problem. If this kernel ever becomes a significant bottleneck we can revisit it. I would also have to have 2 separate paths as cutorch is still supporting fermi.

csarofeen:
latency on a 32x32 reduction is ~6us, highest bandwidth is 334GB/s 16384 x 2048, Titan X pascal. I don't think it's worth further optimizing unless it becomes a bottleneck.

lib/THC/THCReduce.cuh
apaszke: I'm not concerned about divergence, the only benefit from what I suggested would be reducing the number of unnecessary comparisons (and arguably slightly clearer code).

csarofeen: I've personally found this type of unrolling to be easier to understand. Latency of global loads will overshadow comparisons, they won't be significant for performance in these cases. We're down to personal preferences.

soumith:
is this good to go @csarofeen ?

csarofeen:
@soumith Yes, it is good to go.

PR #774, opened 2017-06-06
Support more than 8 GPUs with P2P in groups

szagoruyko:
If a machine has more than 8 cards and all of them support P2P, torch will fail with:

```
RuntimeError: cuda runtime error (60) : peer mapping resources exhausted at /py/conda-bld/pytorch_1493676237139/work/torch/lib/THC/THCGeneral.c:224
```

That's because it is not possible to enable p2p for more than 8 GPUs. I propose to do what caffe2 does, enabling p2p in groups of 8.

https://github.com/caffe2/caffe2/blob/772c23e4050847b3feef8625fc6b3a630cf7c8e4/caffe2/core/context_gpu.cu#L141

cc: @ngimel

### DIFF
--- a/lib/THC/THCGeneral.c
+++ b/lib/THC/THCGeneral.c
@@ -16,6 +16,10 @@
  * above, whichever is greater.*/
 #define MIN_GLOBAL_SCRATCH_SPACE_PER_DEVICE 32768 * sizeof(float)
 
+/* Maximum number of P2P connections (if there are more than 9 then P2P is
+ * enabled in groups of 8). */
+#define THC_CUDA_MAX_PEER_SIZE 8
+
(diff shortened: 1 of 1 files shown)


soumith:
looks good :)
thanks.

ngimel:
One could imagine some exotic configurations where this would fail whereas previous default handling would work (e.g. 10-GPU system with 2 first GPUs peered, and next 8 peered). Alternative is not to break on "mapping resources exausted" error but rather set p2pAccessEnabled array to 0.

soumith:
in the exotic system case, what would happen is that 6 of the next 8 GPUs will be separated from the last two. But it wont hard-fail, instead will do unpeered copies, which will be slightly slower.

PR #783, opened 2017-06-21
Fp16 fixes for CUDA 9

csarofeen:
fp16 fixes for changes that are coming in CUDA 9

### DIFF
--- a/lib/THC/THCAtomics.cuh
+++ b/lib/THC/THCAtomics.cuh
@@ -110,7 +110,7 @@ static inline  __device__ void atomicAdd(half *address, half val) {
     __half_raw hsum;
     hsum.x = (size_t)address & 2 ? (old >> 16) : (old & 0xffff);
     half tmpres = THCNumerics<half>::add(hsum, val);
-    hsum.x = reinterpret_cast<unsigned short>(tmpres);
+    hsum = __half_raw(tmpres);
 #endif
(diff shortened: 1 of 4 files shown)


lib/THC/THCAtomics.cuh
```diff
@@ -102,9 +102,16 @@ static inline  __device__ void atomicAdd(half *address, half val) {
...
     assumed = old;
+#if CUDA_VERSION < 9000
     half hsum;
     hsum.x = (size_t)address & 2 ? (old >> 16) : (old & 0xffff);
     hsum = THCNumerics<half>::add(hsum, val);
+#else
+    __half_raw hsum;
+    hsum.x = (size_t)address & 2 ? (old >> 16) : (old & 0xffff);
+    half tmpres = THCNumerics<half>::add(hsum, val);
+    hsum.x = *(unsigned short*)&tmpres;
```
apaszke: while this probably works, technically it's UB (because of the strict aliasing rule). It would be better to change it to `hsum.x = reinterpret_cast<unsigned short>(tmpres);`. Same for all casts below

csarofeen:
Don't merge this yet.

csarofeen:
Closing for now

csarofeen:
A __half_raw(half) constructor was made for CUDA 9, now we don't need type punning or reinterpret cast.

PR #784, opened 2017-06-26
Alias multinomial sampling in Cuda

amartya18x:
Cuda version of [Alias Multinomial PR in torch7](https://github.com/torch/torch7/pull/1046)

PR #786, opened 2017-06-28
making Multinomial sampling slightly faster

amartya18x:
This makes multinomial sampling slightly faster by generating a `sample` sized array of uniformly sampled random numbers outside the kernel and passing it to the kernel. The different distributions in the kernel can use the same array of uniformly sampled numbers as they are independant distributions.

I have addded a file called `test/multinomial.lua` to do some benchmarking. I can remove it later if the PR is accepted.


amartya18x:
The __NEW__ block shows the time for the current code and the __old__ block shows it for the previous code.

amartya18x:
@pavanky Could you have a look at this too ?

lib/THC/generic/THCTensorRandom.cu
```diff
@@ -202,13 +202,21 @@ THC_API void THCTensor_(multinomial)(struct THCState *state,
...
-          gen->gen_states,
-          n_sample,
-          THCudaLongTensor_data(state, self),
-          numDist, numCategories,
-          THCTensor_(data)(state, prefixSum));
+           THCTensor_(data)(state, uniform_idx),
+	   n_sample,
+	   THCudaLongTensor_data(state, self),
+	   numDist, numCategories,
+	   THCTensor_(data)(state, prefixSum));
```
pavanky: All the lines are not aligned.

amartya18x:
@pavanky But I think generating just an array (the size of `n_sample` instead of `num_dist * n_sample`) is a good improvement. Could it be a good update for now ? Maybe it can be improved again later?

soumith:
@amartya18x what speedups are you seeing? Can you give the output of the benchmark and specify what GPU you ran them on?

wickedfoo:
Yeah, I'm not convinced.

Not only does it require allocation and deallocation (which is bad if one is not running with the caching allocator), it's replacing in-kernel generation with extra global memory reads and writes, which together with the memory allocator stuff may be on par with what was there before. Do you have kernel timings?

lib/THC/THCTensorRandom.cuh
```diff
@@ -292,7 +291,7 @@ sampleMultinomialWithoutReplacement(curandStateMtgp32* state,
 
     // All threads must participate in this
     T r = ScalarConvert<float, T>::to(curand_uniform(&state[blockIdx.x]));
```
wickedfoo: plus, this wasn't converted

test/multinomial.lua
```diff
@@ -0,0 +1,52 @@
...
+local function checkMultinomial()
+   local n_class = {10, 100, 1000}
+   local n_sample = {10, 100, 1000, 10000}
+   local n_dist = 100
+   for _, curr_n_class in pairs(n_class) do
+      for _, curr_n_sample in pairs(n_sample) do
+	 print("")
+	 print("Benchmarking multinomial with "..curr_n_class.." classes and "..curr_n_sample.." samples")
+	 torch.seed()
+	 local probs = torch.CudaDoubleTensor(n_dist, curr_n_class):uniform(0,1)
```
wickedfoo: I'm not sure that benchmarking using float64 is useful, since almost all work done with Torch is in float32. Furthermore, this will have very skewed results on different GPUs due to lack of float64 ALUs.

test/multinomial.lua
```diff
@@ -0,0 +1,52 @@
...
+	 local a = torch.Timer()
+	 local cold_time = a:time().real
+	 a:reset()
+	 cutorch.synchronize()
+	 a:reset()
+	 for i = 1,10 do
+	    torch.multinomial(probs, curr_n_sample, true)
+	    cutorch.synchronize()
```
wickedfoo: why are you synchronizing every time through for the benchmark? one should only synchronize at the beginning and the end.

lib/THC/THCTensorRandom.cuh
amartya18x: Yeah, I didn't do it for this. I just tried it out for the other one. If the idea convinced everyone, I could do it for the other one too.

amartya18x:
@soumith I had the benchmarking here in a comment showing the speedups. I think the rebasing messed it up. Let me check

test/multinomial.lua
amartya18x: Do you suggest using `torch.FloatTensor` ?

pavanky:
@wickedfoo the problem was the previous kernel was generating more random numbers than needed. This PR fixes that issue by creating the required number of random numbers outside.

That said, this isn't the most optimal fix. We should be generating the necessary random numbers within the kernel.

wickedfoo:
Is generating more random numbers than needed really that much of an issue? If we're talking microseconds here, this isn't that big of a deal in my opinion.

The RNG being used requires the entire block to update the state at once, although you are right that only one value per warp is being used.

The alternative would be to see if warp divergence for the binary search part isn't too bad, in which case every thread would use the value that it generated locally.

pavanky:
@wickedfoo Another alternative would be to perform parallel linear search instead of a binary search. 

Either way the improvements here would be minimal. The more substantial improvements we are seeing come from here: https://github.com/torch/cutorch/pull/784

pavanky:
> The RNG being used requires the entire block to update the state at once, although you are right that only one value per warp is being used.

AFAIK this is only a limitation of MTGP32 generator. This is also a bit slower than other generators i CURAND. Philox may even be faster. Any reason to stick with MTGP32 ?

pavanky:
Nvm my previous comment, I didnt realize cutorch only supports this generator.

